PR #3, opened 2017-05-23
Moar chaos (and fix a tyop, so less chaos)

armcknight:


PlasmaPower:
:-1: there should be an equal chance of a decrease, currently this helps everything

Ameliorate:
:-1: See above.

Cubox:
👎

LaurenceOfAustralia:
👎

PR #6, opened 2017-05-23
Use a shell script to restart ChaosBot instead of python

Ameliorate:
This allows things such as redirecting stdout and stderr to a file, as well as making certain shell actions easier.

In the future, https://docs.python.org/2/library/simplehttpserver.html or the such can be used to host the stdout of the bot for easier debugging. 

Do also note that this is mostly untested, and should be reviewed before being merged.

### DIFF
--- a/chaos.py
+++ b/chaos.py
@@ -19,7 +19,7 @@ from github_api import exceptions as gh_exc
 import patch
 
 
-if !os.path.isfile("usestartscript"):
+if not os.path.isfile("usestartscript"):
     open("usestartupscript", 'a').close()
     subprocess.call("start.sh", shell=True)
 
--- a/chaos.py
+++ b/chaos.py
@@ -19,6 +19,11 @@ from github_api import exceptions as gh_exc
 import patch
 
 
+if !os.path.isfile("usestartscript"):
+    open("usestartupscript", 'a').close()
+    subprocess.call("start.sh", shell=True)
+
+
 THIS_DIR = dirname(abspath(__file__))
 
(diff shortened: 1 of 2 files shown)


trentstollery:
👍

rtzoeller (approved):

rtzoeller:
👍

Ameliorate:
`exec > >(tee -a stdout.txt)` could be used to redirect stdout to a file by putting it on the first line of start.sh, for when we want to do that.

amoffat:
I should elaborate that chaosbot is currently running under [supervisor](http://supervisord.org/) on its server.  This is its config file:

```
[program:chaos]
command=/root/.virtualenvs/chaos/bin/python chaos.py
directory=/root/workspace/Chaos
user=root
```

stderr and stdout are currently logged to `/var/log/supervisor/chaos-std*`.  Personally, I think it's good to leave chaos under the watchful eye of supervisor, so :-1: from me

Ameliorate:
Thanks for the feedback, I'll use that file when I get on my desktop and can write some more advanced code.

chaos.py
```diff
@@ -19,6 +19,11 @@
 import patch
 
 
+if not os.path.isfile("usestartscript"):
+    open("usestartupscript", 'a').close()
```
eighthjouster, line 23: Shouldn't it be `usestartscript` like in the previous line?

Ameliorate: Nice catch, but this probably isn't getting approved so I won't bother to fix that.

I'll close the PR since it looks so.

PR #7, opened 2017-05-23
Revert "Comment on approved prs with the linux distro" now that we know the distro

Ameliorate:
Reverts chaosbot/chaos#2

We now know the distro is Ubuntu 14.04 trusty.


sosodev:
👍 but we should add to the readme probably

PR #10, opened 2017-05-23
Adding a simple http server for current directory

richardanaya:
Here's a simple server, however, I don't know how to best call this process cleanly from the main chaos.py.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1,0 +1,5 @@
+<html>
+<body>
+Hello World!
+</body>
+</html>
--- a/server.py
+++ b/server.py
@@ -1,0 +1,7 @@
+#create a simple http server
+import http.server
+import socketserver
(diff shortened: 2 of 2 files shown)


Ameliorate:
This is a very bad idea. In the current directory is the `github_pat.secret` file, which contains the github secret. Having a http server serving the CD would allow this file to be viewed by anyone, and someone will steal the account and derail the bot. :-1:

amoffat:
Bingo @Ameliorate.  If this was changed to prevent serving that file, or if we could confirm that if that file was moved up to the parent directory, `SimpleHTTPRequestHandler` would not be able to access it, then there's no issue

richardanaya:
oh, thanks! sorry about that, still figuring out the environment

PR #11, opened 2017-05-23
shim for moving the github_pat.secret file into safer place

amoffat:
then it can be moved to /etc, and the shim removed.  would allow for features like #10 which serves files in the local directory

### DIFF
--- a/settings.py
+++ b/settings.py
@@ -1,3 +1,6 @@
+from os.path import exists, abspath, dirname, join
+THIS_DIR = dirname(abspath(__file__))
+
 # this is a personal access token used by chaosbot to perform merges and other
 # api requests.  it is a secret, and lives on the server, but since chaosbot has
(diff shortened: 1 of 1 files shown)


AuroraLS3:
👍

patricknelson:
Why not serve from a sub directory where files above that aren't accessible?

eighthjouster:
👍

amoffat:
@patricknelson that works too.  i mainly just want `github_pat.secret` out of potential harms way

patricknelson:
Just saying 😄 Effectively there's no difference between `.` or `/etc`  to `chaosbot` (being it runs as root) if the public directory is lower down. I think the argument for moving it entirely out of the repo root is so unsuspecting PR contributors don't accidentally expose it (devil's advocate here), so that's my argument for moving it out.

PR #13, opened 2017-05-23
Ｃｒｅａｔｉｎｇ　ｓｅｒｖｅｒ　ａｎｄ　ａ　ｗａｙ　ｔｏ　ｋｉｌｌ　ｉｔ

richardanaya:


server/server.py
```diff
@@ -0,0 +1,17 @@
+import http.server
+import socketserver
+
+#set the process name to "chaos_server" so we can easily kill it with "pkill chaos_server"
+def set_proc_name(newname):
+    from ctypes import cdll, byref, create_string_buffer
+    libc = cdll.LoadLibrary('libc.so.6')
```
agargiulo, line 7: Why not put the lib name in `settings.py`? Or are we just assuming it's not changing because 14.04?

agargiulo: Otherwise 👍

patricknelson:
I agree, should be in `settings.py`. This is utter chaos!

agargiulo:
Best part is without a copy of 14.04 installed myself I can't test that it's even the correct lib, plus the internet warns against this way of things.

richardanaya:
Yah =/ I was trying to test this in a recent ubuntu vm but not quite the same. I'm curious how you all think I should reference the settings file since its in a parent directory of the server.py file that gets executed.

agargiulo:
There is a way to set the Document Root of the server without running it from the same directory as `server.py`. I just haven't touched this in long enough to remember

amoffat:
I'd love to clone the fork to try it out, but I'm seeing `unknown repository` for the head.  Any chance you can get a copy of the code up somewhere @richardanaya ?

agargiulo:
It looks like the "unknown repo" thing has been around for years now. Great.. https://github.com/isaacs/github/issues/168
It means he deleted the fork/branch/etc if I'm reading this correctly.

amoffat:
@agargiulo makes sense

jabbink:
RIP

amoffat:
i believe there is now some `TIME-WAIT` action happening on the socket now.  pkill can kill the process, but it can't acquire the socket while it is in `TIME-WAIT` shutdown state.  the result is that whenever chaosbot restarts, it alternates between having a working webserver and not

PR #15, opened 2017-05-23
Ｕｐｄａｔｅ　ＲＥＡＤＭＥ　ｗｉｔｈ　ＯＳ

richardanaya:


agargiulo:
What are you doing to all your PR titles and why?

richardanaya:
Chaos

AndreiRailean (approved):

agargiulo:
Is that some sort of weird web font junk?

demiters:
Good lord what's with the title font, I'm inclined to downvote all of these for that.

nurupo:
It's the same font as always. Only GitHub can change the font of the issue title, there is no way for a GitHub user to change the font to, say, Comic Sans, or some [web font](https://fonts.google.com/). The letters used in the font, however, are not your regular Latin letters, those are full-width Latin letters. [Here is the Unicode range for them](http://unicode.org/charts/PDF/UFF00.pdf).

agargiulo:
Oh that's more vile. (And more what I was thinking but making jokes about web fonts is fun)

demiters:
Accepting these chaotic titles sets a bad precedent! :confused:

ashleysommer:
Also not liking the full-width characters in the title.

ghost:
I'm all for nominating @demiters as wet blanket of the year 2017

PR #16, opened 2017-05-23
There was an unfortunate lack of cats.

hawkfalcon:
This is the internet, after all.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,6 +7,8 @@
 ChaosBot is a social coding experiment to see what happens when the absolute
 direction of a software project is turned over to the open source community.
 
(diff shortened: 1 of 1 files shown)


agargiulo:
👎 No gifs in readmes ⛔️

hawkfalcon:
Would you prefer sourcing from http://thecatapi.com/api/images/get?format=src&type=png ?

agargiulo:
Eh, it would be better but still kinda obnoxious more than chaotic. Is there at least a way to make them not MASSIVE?

hawkfalcon:
Yep, looks like there is a size parameter which takes small, medium, or large. Will create new PR.

PR #17, opened 2017-05-23
There is an unfortunate lack of cats (improved)

hawkfalcon:
This is the internet, after all.

Improved based on feedback from https://github.com/chaosbot/chaos/pull/16


demiters:
:smirk_cat:

kdavis:
Still think there's not enough cats...

froesecom:
The fact that this pull request got up-voted enough to get merged, clearly demonstrates the tragedy of the commons.

PR #19, opened 2017-05-23
reddit chaos bot

bradmonk:
I made a simple reddit bot using PRAW that will scrape a reddit thread
where people are talking about chaosbot and print the newest comment. I
think it would be fun for someone to expand this code to have chaosbot
scrape for comments about chaos, and maybe leave some sort of reply.

### DIFF
--- a/redditchaosbot.py
+++ b/redditchaosbot.py
@@ -1,0 +1,35 @@
+import praw
+from praw.models import MoreComments
+import pprint
+
+
+'''Authenticated instance of Reddit'''
+# --------------------------------------------------------------------
+# https://praw.readthedocs.io/en/latest/getting_started/installation.html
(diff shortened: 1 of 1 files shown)


amoffat:
Just saw this https://www.reddit.com/r/programming/comments/6criij/chaos_a_social_coding_experiment_that/dhx7t2g/

amoffat:
Also if it's a work in progress, add "WIP" to your title so the bot doesn't consider it before its done

bradmonk:
Yeah kinda funny whoever did that. I created a throw-away for chaosbot. If people abuse that account the only option would be to create another in non-authenticated mode (no password provided). But then the bot couldn't comment.

Isn't this entire idea a constant WIP? The bot is fully functional so I guess it can be considered complete. It's up to others to make changes.

PR #25, opened 2017-05-23
Create .hgignore

sugar700:
Mercurial is a superior VCS system nobody uses, let's make it possible to change this repository into Mercurial repository. Yay!


kdavis:
Our chaos git overlord does not agree!
👎

PR #27, opened 2017-05-23
Provide a way to automatically install new requirements

OpenNingia:
We need more shiny Python modulesssss :snake:

### DIFF
--- a/chaos.py
+++ b/chaos.py
@@ -41,7 +41,9 @@ def restart_self():
     """ restart our process """
     os.execl(sys.executable, sys.executable, *sys.argv)
 
-
+def install_requirements():
+    """install or update requirements"""
+    os.system("pip install -r requirements.txt")
 
 if __name__ == "__main__":
     logging.info("starting up and entering event loop")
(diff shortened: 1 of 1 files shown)


chaos.py
```diff
@@ -41,7 +41,9 @@ def restart_self():
     """ restart our process """
     os.execl(sys.executable, sys.executable, *sys.argv)
 
-
+def install_requirements():
+    """install or update requirements"""
+    os.system("pip install -r requirements.txt")
```
amoffat, line 46: I believe this will execute the system's pip, not the virtualenv's pip.  Strip out the virtualenv's bin directory from `sys.executable` to get that pip.

OpenNingia: I don't think this is the case. Since [system()](http://man7.org/linux/man-pages/man3/system.3.html) fork the calling process the child share the environment.

Example:

```
openningia:~/venv_test $ virtualenv test
New python executable in test/bin/python
Installing setuptools, pip...done.
openningia:~/venv_test $ source test/
bin/     include/ lib/     local/   
openningia:~/venv_test $ source test/bin/activate
(test)openningia:~/venv_test $ python
Python 2.7.6 (default, Oct 26 2016, 20:30:19) 
[GCC 4.8.4] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import os
>>> os.system("pip --version")
pip 1.5.4 from /home/ubuntu/venv_test/test/local/lib/python2.7/site-packages (python 2.7)
0
>>> 
```

amoffat: aha, cool :+1:

MCSH (approved):

DasSkelett:
👍

fokoepxg:
:+1:

PR #34, opened 2017-05-23
Creating a "stats page" on http.

hank-der-hafenarbeiter:
start of a flask to show some stats on http.

### DIFF
--- a/chaosstats.py
+++ b/chaosstats.py
@@ -1,0 +1,27 @@
+import os
+import sys
+import json
+import inspect
+
+try:
+    from flask import Flask 
+except ImportError:
+    os.system("pip install Flask")
+    from flask import Flask 
+app = Flask(__name__)
+
+config_file_path = ""
+
+@app.route("/")
+def showChaosStats():
+    try:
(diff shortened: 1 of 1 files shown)


amoffat:
webservers everywhere

:+1:

ghost:
consider using a simpler httpd and static pages instead?

maybe use stagit? http://git.2f30.org/stagit/log.html

hank-der-hafenarbeiter:
I avoided httpd because i wouldn't want to deal with having to kill my processes if/when the bot dies/restarts. In my opinion this approach is easier because the sever is killed when the script stops. The server in the server file uses httpd and solved it with the application name. maybe that is a better approach.

ghost:
Agreed, using external tools would lead to an unmaintanable project.
I guess it leaves us with to choices:
Don't use any external tools or use something like ansible to manage everything.

xylix (approved):

hank-der-hafenarbeiter:
I think it makes more sense to do one http server for everything and since the other server is already up and running i'm retracting this PR.

PR #42, opened 2017-05-23
For a true democracy all valid votes must be equal

br0kend:
In the interest of fairness and correctness, a truly democratic voting system should be used. No voter shall have more power than another one. The principles of equality should reflect the will of the people.

### DIFF
--- a/github_api/voting.py
+++ b/github_api/voting.py
@@ -118,15 +118,8 @@ def get_vote_weight(api, username):
     created = arrow.get(user["created_at"])
     age = (now - created).total_seconds()
     old_enough_to_vote = age >= settings.MIN_VOTER_AGE
-    age_weight = 1.0 if old_enough_to_vote else 0.0
+    weight = 1.0 if old_enough_to_vote else 0.0
 
-    # here we use some basic social proof to weight their vote.  the theory here
-    # is that a user's followers has a log relationship to their coding
-    # judgement?  there's probably something better to use here
-    followers = user["followers"]
-    social_weight = log(followers + 1, settings.FOLLOWER_LOG_BASE)
-
-    weight = age_weight * social_weight
     return weight
 
 


4a616d6573205265696c6c79:
👍

leo-p:
👍

notjonmadden:
👍

jasondecastro:
👍

qn0x:
:+1:

PlasmaPower:
`FOLLOWER_LOG_BASE` should be removed from the config if this change goes through.

GUIpsp:
This shouldn't be merged because then anyone can make sockpuppets and manipulate the system.

OpenNingia:
@GUIpsp I don't follow; this PR only changes the weight of the votes, it doesn't remove them.

pchauncey (approved):

qn0x:
@OpenNingia I think what @GUIpsp meant is that everyone can create new accounts and thus make "their" option weigh more.

OpenNingia:
@qn0x They would have to wait a month tho, because that requirement was not lifted with this PR.

jannemann:
👍

github_api/voting.py
```diff
@@ -118,15 +118,8 @@ def get_vote_weight(api, username):
     created = arrow.get(user["created_at"])
     age = (now - created).total_seconds()
     old_enough_to_vote = age >= settings.MIN_VOTER_AGE
-    age_weight = 1.0 if old_enough_to_vote else 0.0
+    weight = 1.0 if old_enough_to_vote else 0.0
```
rec, line 121: Why not just `return float(age >= settings.MIN_VOTER_AGE)`?  Python guarantees that `True` and `False` always convert to `1.0` and `0.0`.

MarHoff:
👍

github_api/voting.py
netgusto: Because it keeps the intent clearer, while sticking to the original API (float) ?

pchauncey (changes requested):
Please also update the README.md to reflect this change?

lcpz:
Your PR is too naive, this is not a perfect world.

You should implement something smart and hopefully not to big to prevent malicious people from doing as they please.

The log weighting is simple and it works, so I'm downvoting for now.

rtzoeller:
👍

Nonepse:
The debate here is to know what is better between a democracy and a meritocracy. I am in favor of meritocracy, so 👎

dskoda1:
👍

Fire-Hound:
👎

noway:
Power to the people! 👍

stoicio:
:+1:

lcpz:
Did you read the PR? I believe this is not power to the people, which I'd want as well.

It's power to anything.

Khoyo:
This is *chaos*bot, so we should promote chaos. 👍

svanschooten:
While I do believe in the good of people and that they will not abuse a system like this, blockchain technology has been struggling with this problem for a while now. In that scope its called a [sybil attack](https://en.wikipedia.org/wiki/Sybil_attack) and there are methods to cope with this. The age method is quite simple and effective.

maxcaron:
👍

jabbink:
ITT: people without (many) followers.

joelebeau1:
👍

github_api/voting.py
yozel: Explicit is better than implicit.

MegaApuTurkUltra:
👍

b0o:
Follower counts are just as easy to inflate as vote counts. Just have all of your fake accounts follow each other. 

Perhaps a better solution may be to hard-code usernames of trusted contributors who have some sort of regulated veto power but still otherwise have normal vote weight. 

This could help prevent blatantly malicious vote manipulation at the very least, while also maintaining a fairer voting system. Maybe these trusted users could also be able to blacklist known bot accounts.

Raane:
👍

aloisdg:
> For a true democracy all valid votes must be equal

I disagree. What about a Liquid/delegative democracy:


1. **Choice of Role**: Each member can choose to take either a passive role as an individual or an active role as a delegate, differentiating this from representative forms in which only specified representatives are allowed. Delegates have further choices as to how active they are and in what areas.
2. **Low Barrier to Participation**: The difficulty and cost of becoming a delegate is small, and in particular does not require political campaigning or winning a competitive election.
3. **Delegated Authority**: Delegates exercise power in organizational processes on behalf of themselves and those individuals who select them as their delegate. Different delegates, therefore, can exercise varying levels of decision power.
4. **Privacy of the Individual**: To avoid social pressures or coercion, all votes made by individuals are private, both from other individuals and from delegates.
5. **Accountability of the delegates**: To ensure the accountability of delegates to their voters and to the community at large, all formal deliberative decisions made by delegates are completely public (or in some forms viewable only to their constituents).
6. **Specialization by Re-Delegation**: Delegates can not only act directly on behalf of individuals as generalists, but through re-delegation they can also act on behalf of each other as specialists.

[More on Wikipedia](https://en.wikipedia.org/wiki/Delegative_democracy)!

scf4:
Oh I love the idea of this whole experiment. 

For those interested in chaos, how about PR which gives significantly more voting power to those who vote yes on the PR?

MegaApuTurkUltra:
> This could help prevent blatantly malicious vote manipulation at the very least, while also maintaining a fairer voting system. Maybe these trusted users could also be able to blacklist known bot accounts.

This is `chaos`bot. We don't care about preventing anything. The true will of the people must prevail.

4a616d6573205265696c6c79:
No gods, no masters.

I am dead against having certain users with more power than others. This isn't an oligarchy.

b0o:
> This is chaosbot. We don't care about preventing anything. The true will of the people must prevail.

The true will of the people, or of the one person with the biggest collection of fake github accounts? 😏

ghost:
How about applying not a log function, but a sigmoid?

We can scale it by a certain constant (say, 5 or 10) to scale the effect of the number of followers.

e.g., using 2*Sigmoid(n/5). 

| Followers       | Votes assigned   |
|:-------------:|:-------------:|
| 0 | 1 |
| 5 | 1.46 |
| 10 | 1.76 |
| 20 | 1.96 |
| 30 | 2.00 |

The influence of "power users" quickly levels off.

Any fake accounts will need to gather followers to effectively take over the system, and it makes it harder for them to influence the system compared to when they already have the weight of real users. Maybe we can also include commits, watching, stars, etc. in the calculation, but this will complicate things considerably.

This adjustment would be like making something cost $2 vs it being free. You will discourage heavily the amount of dishonest behavior, despite the barrier not actually being that high.

MegaApuTurkUltra:
> The true will of the people, or of the one person with the biggest collection of fake github accounts? 

Check for accounts being X amount of time old and actively contributing on github? Using follower count is completely unnecessary.

Ajedi32 (changes requested):
If this change goes through, the README needs to be updated: https://github.com/chaosbot/chaos#weights-and-thresholds

averaart:
👍

github_api/voting.py
awnumar: I agree with @rec on this.

awnumar:
*"All animals are equal, but some animals are more equal than others."*

👍

lorenzowritescode:
Did you guys seriously not learn anything from Brexit? 👎

awnumar (approved):

0x62 (approved):

bperson:
How about: every account starts with the same voting power (regardless of any other metric). On each merged / rejected PR, whoever voted for the right outcome gets an increase in its voting power?

drusepth:
:+1:

4a616d6573205265696c6c79:
@bperson What is the *right* outcome? Why should people have more voting power because they happened to be on the winning side more often than not?

awnumar:
@bperson nah, then when you have one feature merged, it's easier to get other features merged, and get more weight, and so on. It's an avalanche effect.

bperson:
@4a616d6573205265696c6c79 as you said, there is no "right" or "wrong" outcome :) The only thing you can do is limit or encourage certain behaviours.

All projects like this die out super quickly: encouraging activity and "giving a reward" for it is what we should optimise for imho.

@libeclipse you can limit the voting power (and even "decay" it) if needed. But in the beginning, an "avalanche effect" of activity is something worth encouraging imho, otherwise this leads nowhere :) 

In the meantime, this introduce a nice property / solve an issue that will arise if this project stays active: limiting the power of spammers. As demonstrated on this thread, all metrics that are outside this project's scope can be manipulated (follower count, activity on outside repos ... )

github_api/voting.py
rec: I'm always looking to reduce the number of conditionals, thus the cast, but I can see your point.

At the least, consider getting rid of the antipattern:

    weight = ...
    return weight

which has no documentary value here in a short function called `get_vote_weight`, leading to:

   return 1.0 if age >= settings.MIN_VOTER_AGE else 0.0

Joshfindit:
Avalanches are great...
unless you're in front of them.

github_api/voting.py
yet-another-account, line 121: All votes are equal, but some are more equal than others.

reddraggone9: They're not real people if they're not old enough to vote.

drusepth: Or have ideologies we disagree with.

PR #46, opened 2017-05-23
Living up to the Name

Kyu:
Chaos.


PlasmaPower:
:-1: So this just removes the message about who voted (and what the result was)? We can already find that out by looking at the issue.

Kyu:
I think I fucked up

Issue #18, opened 2017-05-23
What should the voting window be?


amoffat:
Right now it's 1 hour between 10am and 10pm PST, 6 hours otherwise.  Are these values too short or too long?

[Relevant code](https://github.com/chaosbot/chaos/blob/master/github_api/voting.py#L184-L196)

```python
def get_voting_window(now):
    """ returns the current voting window for new PRs.  currently, this biases
    a smaller window for waking hours around the timezone the chaosbot server is
    located in (US West Coast) """
    local = now.to(settings.TIMEZONE)
    lhour = local.hour

    hours = 1
    if lhour <= 10 or lhour >= 22:
        hours = 6

    seconds = hours * 60 * 60 * settings.VOTE_WINDOW_SCALE
    return seconds
```

bradmonk:
If you're finding 1hr is too long, the window could be based on a number of votes (e.g. .5*NumberOfForks)

studiosi:
Take into consideration that some of us want to contribute from different time zones (I am on Eastern European Summer Time now), I think that the right approach to do this is a big window timeframe and some sort of threshold (a PR does not go through unless there is some sort of agreement, whether it is YAY or NAY).

amoffat:
@studiosi what kind of window and threshold are you thinking?

studiosi:
I have to think a little bit about that, but something like one full day for voting on each PR, so everyone can vote and then measure something like how fast the voting has happened (whether it is positive or not), and if the voting is positive or negative with enough speed accept (or reject) the PR before the 24 hours, to increase the number of changes, which is at the end the funny thing about this.

amoffat:
One thing to keep in mind with the longer voting windows would be https://github.com/chaosbot/chaos/pull/26#issue-230631280

studiosi:
A good speed measurement that also can contribute to the democracy here is just keeping a mean of the amount of positive and negative votes per hour per PR and just establish empirically a threshold of speed to just fast-accept or fast-reject.

Regarding #26, I saw it, and I made a proposal here which can help also to end the problem of solving the conflicts.

dskoda1:
Why wouldn't the window just be 24 hours? This way anyone in any time zone has within one day of a or to approve or reject it.

b0o:
Some weird edge cases to consider: 

[...]

PR #51, opened 2017-05-23
Change voting window to 24 hours

golergka:
As per #18

### DIFF
--- a/github_api/voting.py
+++ b/github_api/voting.py
@@ -185,15 +185,8 @@ def friendly_voting_record(votes):
 
 
 def get_voting_window(now):
-    """ returns the current voting window for new PRs.  currently, this biases
-    a smaller window for waking hours around the timezone the chaosbot server is
-    located in (US West Coast) """
-    local = now.to(settings.TIMEZONE)
-    lhour = local.hour
+    """ returns the current voting window for new PRs. 24 hours by default
+    to give the same treatment to users in all timezones"""
 
-    hours = 2
-    if lhour <= 10 or lhour >= 22:
-        hours = 3
-
-    seconds = hours * 60 * 60 * settings.VOTE_WINDOW_SCALE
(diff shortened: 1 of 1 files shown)


hxjuneja:
We need a better solution here, 24 hours will be too much as we will have more contributors

golergka:
As we will have more contributors, we will have more contributors from different time-zones. Have you ever played on MMO server where majority of population were from the other part of the globe and everything interesting happens in the early morning hours? It's not that fun.

dskoda1:
What exactly is wrong with 24 hours? If people are contributing from different time zones across the world this gives everyone a chance to vote on it, can someone that down voted please provide a counter argument?

@hardikj if we have more contributors don't we want to allow everyone to cast their vote? Not limit the # of votes by having it be a couple hours

bperson:
You could also have value in dealing with the chaos left by the "previous timezone" :) + while the project has steam, there is value in having it move fast.

OpenNingia:
24hours is too slow, and could lead the project into starvation. Let's say you need to rebase your pull request, do you need to wait an additional 24 hours to get it merged?

hxjuneja:
@dskoda1 as I said, 24 hours is too much. In my opinion, we can do better, something on lines of having a dynamic window based on factors like 
- Amount of upvotes vs amount of downvotes since the initiation of the pull request (so if a pull request already have 100 upvotes vs 1 downvotes why wait 24 hours?)
- we can also consider the voting in previous timezones.

psykzz:
We should have every 3 hours. Having a time where you wake up and having the bot be EU dominated vs US dominated adds to the fun. Alot can happen in 24 hours, 3 hours at least you can see your results the same day and will keep people interested.

demonspork:
could we calculate the voting window based on a combination of number of votes and size of the PR? Like, if I make a very large PR it starts the timer at 24 hours. If we get a large number of votes, it drops the timer down for each threshold of votes to a minimum of 3 hours?

PR #58, opened 2017-05-23
Update chaos.py

enthus1ast:
moar chaos

### DIFF
--- a/chaos.py
+++ b/chaos.py
@@ -20,7 +20,7 @@ import patch
 
 
 THIS_DIR = dirname(abspath(__file__))
-
+True = False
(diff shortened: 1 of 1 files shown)


chaos.py
```diff
@@ -20,7 +20,7 @@
 
 
 THIS_DIR = dirname(abspath(__file__))
-
+True = False
```
TheUKDave, line 23: This will break in Python 3:
`SyntaxError: can't assign to keyword`

True, False, and None are reserved

enthus1ast: then no moar chaos  : (

Issue #41, opened 2017-05-23
Voting with more emojis


hongaar:
It should be possible to cast votes with more than only 👍 and 👎 . For example, ❤️  and 🎉 could be interpreted as a positive, and 😕 could be interpreted as a negative vote.

This would introduce some bias (as reactions have more positive than negative options) but that may not be a problem, as more merged PRs === more chaos.

Then what about the 😄 reaction?

studiosi:
The psychology of emoji is very weird. For example, some of my friends consider 😄 ironical most of the times, while others interpret it as plain laughter.

One of the most controversial ones is the dancer ( 💃 ) XDDD

jabbink:
:shipit:

DasSkelett:
With 👍 and 👎 you have plain emojis for voting, everybody can understand. More emojis are only confusing I think.

jabbink:
More chaos though.

rtzoeller:
👍🏽

golergka:
I suggest giving ambiguous emojis random meaning. More chaos.

ghost:
:shipit:

tusharmakkar08:
What about 🥇 ?

PR #61, opened 2017-05-23
Voting based on a list of emoji

hongaar:
I'm an absolute newbie to Python, so please let me know if the code should be improved. (I don't want to be responsible for increasing the death counter 😨)

Solves #41 

Note: The list of emojis is a bit verbose, as `emoji.demojize`, reactions and comment body emojis all seem to use slightly different shortcodes.

### DIFF
--- a/github_api/voting.py
+++ b/github_api/voting.py
@@ -1,6 +1,7 @@
 from math import log
 import arrow
 import re
+from emoji import demojize
 
 from . import prs
 from . import comments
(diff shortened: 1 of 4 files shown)


MegaApuTurkUltra:
What if someone reacts with both a positive and negative emoji?

hongaar:
It uses the positive one (that list is iterated first, and if a hit is found, the vote score is returned before iterating the negative list). Not sure how it was originally...

mattherman:
Perhaps we should add the list of positive/negative emojis to the README?

hongaar:
Reactions on this PR are a good use-case 😄

github_api/voting.py
```diff
@@ -160,21 +161,29 @@ def parse_review_for_vote(state):
 
 def parse_reaction_for_vote(body):
     """ turns a comment reaction into a vote, if possible """
-    try:
```
jetpacktuxedo, line 163: I think that try/except was important...

patricknelson: Maybe that was some underhanded python ([a la...](https://en.wikipedia.org/wiki/Underhanded_C_Contest))?

jetpacktuxedo: I'm not sure that anything that happened here qualifies as "rigorous code review". Lol. I'd guess it was more likely to be a genuine mistake than actively malicious.

hongaar: It was important because reactions content can be `+1` or `-1`, but also other string values. Hence the try/except when casting the content to an `int`. This PR changes that to comparing the string value only.

PR #63, opened 2017-05-23
write all cout input to file so that it can be used in the server

hank-der-hafenarbeiter:
write everything from cout into a file so it can easily be used (on the http server for example)

### DIFF
--- a/chaos.py
+++ b/chaos.py
@@ -24,7 +24,8 @@ import patch
 
 THIS_DIR = dirname(abspath(__file__))
 
-logging.basicConfig(level=logging.DEBUG)
+log_file = os.path.join(THIS_DIR, "cerr.txt")
+logging.basicConfig(filename=log_file, level=logging.DEBUG)
 logging.getLogger("requests").propagate = False
 logging.getLogger("sh").propagate = False
 


rudehn:
Why not just configure the logger to output to a file as well?

hank-der-hafenarbeiter:
doesn't the logger output go to cout anyway? so it will should be routed to the file too.

rudehn:
I think it's just cleaner to let the logger handle all program output. That's what its there for after all

miller-time:
what's the reason for using the names "cout"/"cerr" instead of "stdout"/"stderr"?

OpenNingia:
c++ oriented I guess

hank-der-hafenarbeiter:
So now this PR only redirects the logger output to a file called cerr.txt in the directory of the script.

Why cerr? Logger writes to stderr which is cerr in c++ and i didn't really think that it made a difference. I don't know. It's what I've always been using.

MCSH:
The logger should be enough for this, you dont need to reinvent the wheel.
👎

hank-der-hafenarbeiter:
@MCSH look at the last commit. Redirecting the logger to somewhere useful is literally all it does. 2 Lines of code.

MCSH:
In unix every program should be a [filter](https://en.m.wikipedia.org/wiki/Filter_(software)#Unix). 

>>> In Unix and Unix-like operating systems, a filter is a program that gets most of its data from its standard input (the main input stream) and writes its main results to its standard output (the main output stream). Auxiliary input may come from command line flags or configuration files, while auxiliary output may go to standard error. 

This is just bad practice.

PR #64, opened 2017-05-23
Add netdata installation script to monitor server status

danog:
Install https://github.com/firehol/netdata to monitor server status: an http server will be started on port 19999 (http://chaosthebot.com:19999), showing extremely accurate and detailed stats about RAM/CPU/network/Disk usage.
This will come in useful if people start running heavy stuff like bitcoin miners 
oh no wait that could actually happen :(

### DIFF
--- a/install-netdata.sh
+++ b/install-netdata.sh
@@ -1,9 +1,9 @@
 #!/bin/bash
 cd /tmp/
+[ -d netdata ] && exit
 
 curl -Ss 'https://raw.githubusercontent.com/firehol/netdata-demo-site/master/install-required-packages.sh' >/tmp/kickstart.sh && bash /tmp/kickstart.sh -i netdata-all --dont-wait
 
-rm -rf netdata
 git clone https://github.com/firehol/netdata.git --depth=1
 cd netdata
 
--- a/install-netdata.sh
+++ b/install-netdata.sh
@@ -1,0 +1,10 @@
+#!/bin/bash
+cd /tmp/
+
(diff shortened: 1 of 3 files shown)


OpenNingia:
You should check if it's already installed, otherwise the install process is going to be triggered on every restart.

danog:
Done

syntonym:
This will only listen on localhost, won't it?

danog:
No, it will listen on all IP addresses.

install-netdata.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+cd /tmp/
+
+curl -Ss 'https://raw.githubusercontent.com/firehol/netdata-demo-site/master/install-required-packages.sh' >/tmp/kickstart.sh && bash /tmp/kickstart.sh -i netdata-all --dont-wait
+
+rm -rf netdata
+git clone https://github.com/firehol/netdata.git --depth=1
+cd netdata
+
+sudo ./netdata-installer.sh -u --dont-wait
```
reddraggone9: Isn't the bot already running as root?

install-netdata.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+cd /tmp/
+[ -d netdata ] && exit
+
+curl -Ss 'https://raw.githubusercontent.com/firehol/netdata-demo-site/master/install-required-packages.sh' >/tmp/kickstart.sh && bash /tmp/kickstart.sh -i netdata-all --dont-wait
+
+git clone https://github.com/firehol/netdata.git --depth=1
+cd netdata
+
+sudo ./netdata-installer.sh -u --dont-wait
```
reddraggone9, line 10: Isn't the bot already running as root?

PR #68, opened 2017-05-23
html index  updates

jppope:
Starting to work on the index.html page. partially a test to see how this works


jppope:
- Updated the index file to remove bootstrap in favor of a framework based on flexbox. 
- merged in other index changes

server/index.html
```diff
@@ -1,53 +1,40 @@
-<!DOCTYPE html>
```
rhengles, line 1: Why?

PR #71, opened 2017-05-23
Update README.md - added todo list

beenotung:
added todo list;
suggest to add sample use cases;


sugar700 (changes requested):
Wouldn't it be easier to just put use case examples, instead of TODO?

beenotung:
@xfix because I've no idea how to use this repo as a library / runtime

PR #72, opened 2017-05-23
Implemented a function that let's you call a function only once even across restarts

hank-der-hafenarbeiter:
added a utils directory with a utils.py. Within I implmented callOnce(func).

When a function is passed to it, it will only execute it if a function with that name has never been executed before.

The benifit is that if your code need to set something up you can just put that in a function (foo()) and you can put callOnce(foo) in your script. When your PR is merged this setup function will be called. When the bot restarts callOnce(foo) will **not** execute foo() again. Your setup is done only the first time around.

If you need to set up directories or download packages, you can put it into a function and make a callOnce() call at the beginning of your script. If the function has been called before (during the current execution or any previous run) it won't be called again.


Ajedi32 (commented):
Are you sure this will work? Won't `called_functions.json` get replaced every time chaosbot is re-deployed?

patricknelson:
Well, apparently @chaosbot seems to think it'll work just fine.

hank-der-hafenarbeiter:
It reads the file into the set and at the end it overwrites the file by dumping the set into it. Pretty inefficient, but it should work.

Ajedi32:
@hank-der-hafenarbeiter Yeah, but "the file" is in source control. If the code that deploys new versions of the bot is doing its job properly, the contents of that file will be reset to `[""]` every time new code gets deployed (because that's what the master branch of this repository says that file is supposed to contain).

PR #49, opened 2017-05-23
Organize startup mechanism

PlasmaPower:
By placing everything in a directory (`startup.d`), we can eliminate a lot of merge conflicts and help streamline the process. This is essential if we want to expand chaosbot's functionality in the future.

PR #73, opened 2017-05-23
Organize startup mechanism (#49 but with merge error fixed)

sugar700:
I want this to be included, but it doesn't work, so here is a fixed version. Of course, the author submitted the fixed version, but the votes are void, and you cannot revote, so...?

### DIFF
--- a/chaos.py
+++ b/chaos.py
@@ -1,7 +1,6 @@
 import time
 import os
 import sys
-import sh
 from os.path import dirname, abspath, join
 import logging
 import threading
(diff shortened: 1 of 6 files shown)


PlasmaPower:
> and the votes are void

Oh yeah, I forgot about that. I'm closing my original PR. :+1:

reddraggone9:
You technically can revote by removing your vote and adding it back since [reactions are counted from when they were created][0], but since most people probably won't do that this is probably the best way to go about it.

[0]: https://github.com/chaosbot/chaos/blob/f30a329cdb0da09b9b9cbf1e5500bd5be54fc79f/github_api/prs.py#L159-L166

PR #79, opened 2017-05-23
Now uses all of the normal and offensive fortunes from debian

jetpacktuxedo:
I wanted more fortunes. 
108458 insertions(+), 6995 deletions(-)
Now there are a lot of fortunes.


Ajedi32:
This makes the file 4.16 MB. Not huge, but certainly large enough so as to be difficult to review properly.

jetpacktuxedo:
I don't disagree. Debian solves this buy splitting it out categorically into a bunch of smaller files that are each easier to review. I didn't want to spend a bunch of time editing the actual python just to get a longer list though, so I just cat'd them all together.

PR #48, opened 2017-05-23
Need more chaos

Zidail:
1pk557

PR #90, opened 2017-05-23
Banning users

MCSH:
It might be a good idea to ignore prs from users wjo are banned.

Ideas for further improvement:
* add a poll method to ban a user (using issues preferably)
* deny prs from banned users

### DIFF
--- a/github_api/prs.py
+++ b/github_api/prs.py
@@ -100,6 +100,7 @@ def get_ready_prs(api, urn, window):
     open_prs = get_open_prs(api, urn)
     for pr in open_prs:
         pr_num = pr["number"]
+        user = pr["user"]["id"]
 
         now = arrow.utcnow()
         updated = get_pr_last_updated(pr)
@@ -113,7 +114,7 @@ def get_ready_prs(api, urn, window):
 
         is_wip = "WIP" in pr["title"]
 
-        if not is_wip and delta > window:
+        if not is_wip and delta > window and not user in settings.BANNED:
             # we check if its mergeable if its outside the voting window,
             # because there seems to be a race where a freshly-created PR exists
             # in the paginated list of PRs, but 404s when trying to fetch it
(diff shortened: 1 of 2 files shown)


romgrk:
Ban Zidail :D (The guy from https://github.com/chaosbot/chaos/pull/48)

rudehn:
An alternative, less harsh, approach would be to at least have chaosbot comment on PRs from people with previous PRs deemed really bad by the community.

Could be as simple as "Hey this guy might be malicious/spam".

This gives a heads up for everybody

MCSH:
@rudehn that sounds interesting. I'll work on it later.

MINIMAN10000:
I prefer rudehn's idea for chaosbot leaving a comment on PRs

I think part of the fun in chaosbot is everyone has a vote. Made me happy when I read that democracy vote got accepted. ( Removal of following bonus )

MCSH:
@MINIMAN10000 interestingly the pr got reverted due to server error.

f2cd29ea15de881ab3377cad5519e0c2dcfe00ac

Coincidence?

github_api/prs.py
```diff
@@ -100,6 +100,7 @@ def get_ready_prs(api, urn, window):
     open_prs = get_open_prs(api, urn)
     for pr in open_prs:
         pr_num = pr["number"]
+        user = pr["user"]["id"]
```
amoffat, line 103: use "login" instead of "id"... it's pretty tricky to find your id

amoffat (changes requested):

github_api/prs.py
MCSH: You can change login

amoffat: oh wow, did not know

amoffat:
If anyone is wondering why this got rejected, with #42 put in place, and with more yays than nays, the answer is [here](https://github.com/chaosbot/chaos/blob/master/patch.py#L36).  User vote weights are cached for a day.

PR #100, opened 2017-05-23
inversed voting mechanism

larsnolden:
👎  => positive vote
👍 => negative vote

... chaos :)


EtherTyper:
@larsnolden Even better: automatic acceptance

hanspagel:
👍

reddraggone9:
@EtherTyper We tried that in #48 (which got in on a technicality). It didn't seem like a terribly popular idea, but it was only reverted because it was buggy.

2dvisio:
👍

qwerm:
So does that mean this has 27 downvotes or 27 upvotes?

PR #104, opened 2017-05-23
Added X-Clacks-Overhead

viktorsec:


Redmega:
GNU TERRY PRATCHETT

viktorsec:
@paris-ci http://www.gnuterrypratchett.com/

pchauncey:
I support this overhead

hongaar:
This should be implemented using proper headers!

> **NOTE:**
> Page headers are, by design, sent before the page content is. Using this meta tag will NOT trigger Clacks-Overhead recognition on any of the various browser plug-ins as it is sent AFTER the page headers have already been sent (and therefore cannot be modified). This should be the last resort if you have absolutely no other option to configure your webserver's page headers.

hongaar (changes requested):
This should be implemented using proper headers!

> **NOTE:**
> Page headers are, by design, sent before the page content is. Using this meta tag will NOT trigger Clacks-Overhead recognition on any of the various browser plug-ins as it is sent AFTER the page headers have already been sent (and therefore cannot be modified). This should be the last resort if you have absolutely no other option to configure your webserver's page headers.

PR #85, opened 2017-05-23
Add a tweet on https://twitter.com/Chaos_bot_git when pull request is…

scaule:
This pr help to post a tweet for each pr accepted on this project. 
Post on https://twitter.com/Chaos_bot_git
Feel free to use twitter api :)

PR #105, opened 2017-05-23
API key encryption module

sugar700:
This will allow decrypting stuff like API keys without anyone knowing the API key (other than the person who uploaded those). Depends on secrecy of /etc/privkey.


amoffat:
very clever! :+1:

PlasmaPower:
Was just about to create this myself. :+1:

scaule:
It's a very good Idea, i update #85 with that when pr is accepted

PlasmaPower:
@scaule You might just want to make a new PR. It's hard to get revisions passed.

scaule:
For the moment my PR is ok with credentials in etc .When this PR is accepted, i will create a new PR to use encryption.

PlasmaPower:
@scaule sounds good.

sugar700:
Anyway, gotta go, figure out yourself how to encrypt with encryption Python module. (hint: https://cryptography.io/en/latest/hazmat/primitives/asymmetric/rsa/#encryption)

Decryption is just means of calling decrypt method with binary string, you will get binary string back, but only on chaosbot computer (if you get pubkey.txt from it).

Issue #114, opened 2017-05-23
need a way to configure underling system


g19fanatic:
Need puppet installed!

Work serverless so that any changes to the system need to still be PRs.

g19fanatic:
#115 was merged.

PR #115, opened 2017-05-23
adding masterless puppet

g19fanatic:
Against #114 

Will allow a way to manage the underlying VPS in a configuration managed way without needing to run servers.

### DIFF
--- a/puppet/manifests/sites.pp
+++ b/puppet/manifests/sites.pp
@@ -1,0 +1,3 @@
+node default {
+    class{'chaos_system_packages':}
+}
--- a/puppet/modules/chaos_system_packages/manifests/init.pp
+++ b/puppet/modules/chaos_system_packages/manifests/init.pp
@@ -1,0 +1,13 @@
+class chaos_system_packages {
+	exec { "apt-get update" :
+		command => "/usr/bin/apt-get update"
+	}
+
+	$packages = ['puppet']
+	package {$packages : ensure => "latest", require => Exec['apt-get update']}
+
+	exec { "apt-get upgrade" :
+		command => "/usr/bin/apt-get upgrade -y",
+	        require => Exec['apt-get update']
+    	}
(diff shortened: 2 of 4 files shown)


g19fanatic:
Any discussion on why people think this is a bad idea?

rudehn:
Can you perhaps explain a little more what you want to do?

g19fanatic:
Puppet is an open source configuration management system used to describe system configuration. This will allow us to setup the vps with explicit changes/modules to configure everything you'd be able to configure as a system admin, through chaos voting.

Let's say someone wants to create a front end nginx proxy, that can be done straight from a puppet module in a declarative manner without needing worrying about mistakes in scripts and putting the system in s bad state. Puppet modules allow you to keep the configuration explicitly defined as described.

g19fanatic:
Just want to create the base for the opportunity to install and manage everything else on the vps. Modules can be added as separate PRs. It would be easy to add anything to the vps using a system such as puppet.

Other similar tools are ansible, chef, salt... Etc...

howdoicomputer:
I like the idea but I don't like Puppet so I'm uncertain of whether or not I want to put in an upvote.

patricknelson:
This is actually a good idea, since this'll make it easier to provision the VM this lives on. Also in my local `Vagrantfile` I can ensure my local dev environment is consistent w/ production (see #123)

howdoicomputer:
lol it merges on a tie

amoffat:
@howdoicomputer non-binary vote weights are [still being used](https://github.com/chaosbot/chaos/commit/09117aed7f3b8d52f8856b2447177ab37bf06841)...they were [cached](https://github.com/chaosbot/chaos/blob/master/patch.py#L36) for 1 day to disk.

howdoicomputer:
Thanks for the explanation, @amoffat!

PR #120, opened 2017-05-24
Fallback

jonpepler:
This has become a little more ominous than the original intention, which was to revert the most recent PR if no PR has been approved in the last 24 hours. I found that this was not something that could be done via the API, and so I propose a new consequence for inaction:

**If no PR has been approved in the last 24 hours, Chaos randomly chooses a PR with no conflicts, and merges it in spite of the current vote.** 

The main purpose of this is to encourage progress (I'm afraid of the result of this PR myself), and of course introduce some chaos. This PR ensures that there will always be a discussion to be had. 

_Extra bonus: this PR includes a little class for handling commit objects, might come in handy for others._


MINIMAN10000:
I mean it fits the name chaos. But the thought of adding a commit revert wind up clock makes me uncomfortable.

However I am interested in the idea of being able to revert a commit via api

jonpepler:
@MINIMAN10000 I can't say I'd ever want it to happen myself, but I like the idea of the threat being there 😛

jonpepler:
(updated description & title)

github_api/repos.py
```diff
@@ -4,3 +4,16 @@ def get_num_watchers(api, urn):
...
     return data["watchers_count"]
 
+# returns the latest commit to master if one occurred within the last fallback window
+def get_latest_commit(api, urn):
+    path = "/repos/{urn}/commits".format(urn=urn)
+    data = api("get", path)
+    if(len(data) > 0)
+      # return previous commit sha if it exists
+      if(len(data) >= 1)
+        data[0]['last_commit_sha'] =
```
phil-r: Looks like a mistake

jonpepler: Quite right, thanks for spotting that. I thought I'd removed it.

PR #121, opened 2017-05-24
Restart chaosbot if the server isn't working

gkbrk:


cron/__init__.py
```diff
@@ -2,6 +2,8 @@
 import settings
 
 from .poll_pull_requests import poll_pull_requests as poll_pull_requests
+from .restart_homepage import restart_homepage as restart_homepage
 
 def schedule_jobs():
     schedule.every(settings.PULL_REQUEST_POLLING_INTERVAL_SECONDS).seconds.do(poll_pull_requests)
+    schedule.every(120).seconds.do(restart_homepage)
```
mrhwick, line 9: Maybe should make this magic number `120` a setting?

PR #126, opened 2017-05-24
Readme updates for supervisor

amoffat:
https://github.com/chaosbot/chaos/blob/master/startup.d/90-chaos.sh is really a problem child right now, and I think it was written because people didn't notice the process was already managed under supervisor


caarlos0:
👍

PlasmaPower:
Oh, thanks! Author of `startup.d` here. I just wanted a good way to startup other web servers and stuff, but people have been rolling their own thing for the most part.

PR #123, opened 2017-05-24
👨‍💻 Vagrant environment for local development.

patricknelson:
Vagrant local development environment. Running Ubuntu 14.04 (Trusty) with `python3`, automatically installing requirements and etc. My alternative to the `Dockerfile` request in #116 😄 

**NOTE:** To use this you will still need to manually initialize your github secrets under `/etc` or comment out code referencing github's API's.

See also: https://github.com/chaosbot/chaos/pull/107 (runs in this VM)

PR #129, opened 2017-05-24
use correct pip to install reqs

amoffat:
`pip` called by `30-install-requirements.sh` was calling system pip


PlasmaPower:
Oops again. Why isn't it in the path?

mrhwick (approved):

patricknelson:
Or just call `pip3`? Another reason I think `puppet` (https://github.com/chaosbot/chaos/pull/115 or similar) should be used so there's no mystery as to the system configuration. For my VM #123 I was easily up and running via `apt-get install python3-pip`.   That being said -- wouldn't `pip3 ...` be more elegant?

amoffat:
@PlasmaPower I'm not sure.  When chaosbot restarted and installed new deps, I noticed they were going in system python3.6's lib:

```
/var/log/supervisor/chaos-stdout---supervisor-CkWQoq.log-Requirement already satisfied: emoji==0.4.5 in /usr/local/lib/python3.6/dist-packages (from -r requirements.txt (line 9)
```
Maybe `30-install-requirements.sh` was losing its env from was [how was being executed?](https://github.com/chaosbot/chaos/blob/master/startup.sh#L4)

PlasmaPower:
@ammoffat Nope, it's not loosing it's ENV. It's being executed just like any other program.

amoffat:
@PlasmaPower I'm really confused then.  It was definitely installing to the system python...when the server restarted, it didn't have the dependencies, but when I ran `/usr/bin/python3.6` it did. :thinking:

PlasmaPower:
Odd. Maybe it has to do with how the server restarts itself? My method of execution definitely keeps env:

```
$ export EXAMPLE=1
$ exe="env"
$ "$exe"
(my shell's env)
EXAMPLE=1
```

pchauncey (approved):

amoffat:
@PlasmaPower possibly related, I'm 80% sure the launching of python in [90-chaos.sh](https://github.com/chaosbot/chaos/blob/master/startup.d/90-chaos.sh) is suffering from similar environment issue.  Check this out

```
==> /var/log/supervisor/chaos-stderr---supervisor-lSIrBd.log <==
  File "chaos.py", line 9, in <module>
    import http.server
ImportError: No module named http.server
INFO:chaosbot:starting up and entering event loop
```

These lines come after requirement processing lines.  Btw, you can view log files [here](http://chaosthebot.com:8081/).  The default python on the machine is 2.7, so that would be consistent with not finding `http.server` (since that module doesn't exist in 2.7).  I'm really not sure how the environments are not carrying over into the scripts though, as I ran a similar test to what you pasted and saw similar results.

PR #131, opened 2017-05-24
correct grammar

benharold:
"Their" implies plurality, while the subject is singular.

### DIFF
--- a/github_api/voting.py
+++ b/github_api/voting.py
@@ -32,7 +32,7 @@ def get_votes(api, urn, pr):
         if vote and vote_owner != pr_owner:
             votes[vote_owner] = vote
 
-    # by virtue of creating the PR, the owner casts their vote as 1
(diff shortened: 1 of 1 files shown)


EtherTyper:
@benharold https://en.wikipedia.org/wiki/Singular_they

snood1205:
Furthermore, "his or her" assumes that everyone uses "his" or "her" but that's not accurate.

benharold:
@EtherTyper I have no time to suffer ambiguity. I prefer the concrete.

ECrownofFire:
What if the owner is multiple people working together?

benharold:
@ECrownofFire I believe you mean "What if the owners are multiple people working together?"

ECrownofFire:
"What if the owner is actually a group?" is perfectly valid.

Unless you prefer E-Prime, in which case it'd be "What if the owner instead consists of a group?"

PR #140, opened 2017-05-24
Add a vote counter for previous voters

ECrownofFire:
This creates a `voters.json` file which at the moment just maps from each voter's username to the number of times that they have voted on a PR that chaosbot has considered and accepted or rejected.

The basic idea here is that chaosbot can eventually consider some contributors more or less trustworthy than others, based on the votes they have cast. We can discuss the details of that later.

I strongly recommend against rewarding people for *receiving* votes, as that encourages trivial "safe" content, which lacks chaos.

### DIFF
--- a/cron/poll_pull_requests.py
+++ b/cron/poll_pull_requests.py
@@ -1,5 +1,6 @@
 import arrow
 import logging
+import json
 import os
 from os.path import join, abspath, dirname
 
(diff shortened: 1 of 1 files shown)


amoffat:
Good idea :+1: .  Future iterations can leverage mysql installed on the server, but nobody has braved trying to interact with that yet

hongaar:
Sounds to me like the beginning of omnipresent government surveillance and public manipulation.

PR #138, opened 2017-05-24
Allow the owner of a PR to downvote it

PlasmaPower:
Redo of #77 with merge conflicts fixed.

PR #141, opened 2017-05-24
Encryption, take 2

sugar700:
See #105. Merge error.


lukeramsden:
Maybe try again? @serhalp seems to be downvoting with no good reason.

viktorsec:
@lukeramsden it failed because of #138 

We are no longer able to influence merging by votes.

lukeramsden:
@viktorsec Ahh I forgot about that :P

amoffat:
@xfix re-open when you get a chance, this is a great feature

PlasmaPower:
It's a shame I didn't see this, would have upvoted it (while dictator).

PR #143, opened 2017-05-24
Have some rules!

hongaar:
Provide a (currently empty) rules file, where we can have rules.

Rules could be used as a guideline for evaluating a change proposal, as not to require having the same discussions about changes over and over again. Instead, one can now simply refer to the rules and say: _that's not allowed according to our rules_.

Of course anybody could propose to change the rules.

I propose that when PRs change this new `RULES.md` file, they can change nothing else in the repository, and the PRs should pre prefixed with `[RULES]` for clarity.

So I will make a separate PR for that.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,13 @@
-*.pyc
+### ide
+.idea
+
+### system
 .DS_Store
+
+### local
 .venv
+
+### runtime
+*.pyc
 github_pat.secret
+api_cache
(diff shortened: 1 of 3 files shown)


xyproto:
I think a rules file should be added once someone has a pull request with the first rule to be added instead of a file with no rules (or a rule stating there are no rules).

viktorsec:
@xyproto thats in #144

PR #144, opened 2017-05-24
[RULES] Changing the rules

hongaar:
includes #143


Hello1024:
👎.  Nobody likes rigid rules.   Let the code define the rules.  All text documents are at best guidance or tutorials.

hongaar:
Nobody said _rigid_. But if you're writing code collaboratively, you will need rules.

PR #145, opened 2017-05-24
Fix voting window

ECrownofFire:
It was backwards.

The settings should probably be renamed too, but I don't know what they should be named.


hongaar:
I still think that the voting window should not favour US West Coast but instead be equal for all. But if we have it, it should be done correctly, so have my upvote!

👍

rhengles:
It should be equal for all 👎

PR #146, opened 2017-05-24
Better supervisor config

amoffat:
now that we have http://chaosthebot.com:8081/ for logs, let's make the log files a little more manageable

also, we'll lean on supervisor for restarting things


hongaar:
@amoffat probably try this again after #153 merged

PR #150, opened 2017-05-24
I just MIGHT give you up

ArthurG:


hongaar:
@guoarthur probably try this again after #153 merged

viktorsec:
I think that 75% chance of being redirected is still too damn high.



It would be better to have a — I dunno — 1% chance of being redirected to the rickroll video and in the other 99% not redirect at all.

PR #152, opened 2017-05-24
Voting threshold

hongaar:
There should be a voting threshold (it's actually a threshold on the votes _difference_).

Set to 1% of repo followers (currently, that means the difference between 👍 and 👎 should be at least 12).

Prevents things like #138 and #48.

Maybe we can change this later to use a percentage of active contributors instead of (possibly passive) stargazers? Not sure how to calculate that atm though.

### DIFF
--- a/settings.py
+++ b/settings.py
@@ -43,9 +43,8 @@ FOLLOWER_LOG_BASE = 30
 
 # for a pr to be merged, the vote total must have at least this fraction of the
 # number of watchers in order to pass.  this is to prevent early manipulation of
-# the project by requiring some basic consensus.  i'm not sure its necessary, so
-# it's 0
-MIN_VOTE_WATCHERS = 0
+# the project by requiring some basic consensus.
+MIN_VOTE_WATCHERS = 0.01
 
 # unauthenticated api requests get 60 requests/hr, so we need to get as much
 # data from each request as we can.  apparently 100 is the max number of pages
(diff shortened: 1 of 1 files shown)


rudehn:
I'd maybe do it off a percentage of active participants. Would prevent any deadlocks from occurring

hongaar:
@rudehn I'm all for it. Any ideas how to retrieve that?

ECrownofFire:
With #140 we could check against the number of people who have X or more points.

hongaar:
Oh shit I voted against that 😆 
Maybe we can retrieve a stat on activity on forks in the past x days and use that? Although I think that's quite expensive..

rudehn:
@ECrownofFire sure that's one solution. It would definitely pick up on anybody who has voted for PRs. In the long run even that could get stale if somebody hasn't voted in a long time. Maybe storing some kind of "last active" per user could be useful

ECrownofFire:
@rudehn I was planning to suggest using a certain amount of point decay over time (for each PR) to avoid that.

rudehn:
Sounds good!

rudehn:
I would just like to point out this PR would not pass if this was currently implemented. Perhaps the threshold should be adjusted more?

hongaar:
Good point, yeah I think so. Although I'm leaning towards a conservative threshold, of course that conflicts with creating chaos. Hmmm...

Maybe 0.5% (=7 votes difference atm) is better?

rhengles:
As already mentioned, time decay is a better solution.

PR #153, opened 2017-05-24
🗳 Bring back democracy

hongaar:
Cast your votes, they'll be pointless anyway. 😆

See #138

Our only chance of going back to some form of democracy is if this PR will get the approval of @PlasmaPower.

PR #154, opened 2017-05-24
Fix provisioning for Vagrant at python3-pip installation

ksendor:
Vagrant provision was failing due to lack of `-y` flag for installation of `python3-pip` via `apt-get`.


rudehn:

PR #155, opened 2017-05-24
Update README.md after #138

jakobbotsch:
#138 changed the voting algorithm so only @PlasmaPower's votes count. Updating readme to reflect this.


PlasmaPower:
I'm re-enabling democracy, so you should probably close this PR or it will get merged right after that.

jakobbotsch:
A generous god.

PR #160, opened 2017-05-24
More PR labels

pchauncey:
This PR adds a couple labels to PRs:

- mergeable
- conflicts

### DIFF
--- a/github_api/prs.py
+++ b/github_api/prs.py
@@ -118,12 +118,6 @@ def get_ready_prs(api, urn, window):
         updated = get_pr_last_updated(pr)
         delta = (now - updated).total_seconds()
 
-        # TODO
-        # i would like to do this, but it turns out, the "mergeable" field only
-        # exists on individual prs fetched through the api, not prs in paginated
-        # form
-        #mergeable = pr["mergeable"] is True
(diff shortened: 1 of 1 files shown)


hongaar:
Could this be optimised so that we only add the label if it doesn't already have it?
(considering GH API rate limits might come into play at some point..?)

amoffat:
@pchauncey do you prefer any specific colors for these labels?

pchauncey:
@amoffat You pick!   Something "good" and something "bad"?

pchauncey:
Nice, those are great.  Thanks!

pchauncey:
@hongaar  I like the idea of optimizing, but I'm going to have to save that for another PR.  I need to figure out how to extract the labels from the `pr` object before or during relabel, etc.

hongaar:
@pchauncey would this work?

```python
if "mergeable" not in list(map((lambda x: x["name"]), pr["labels"])):
    label_pr(api, urn, pr, "mergeable")
```

phil-r:
This PR brought a lot of chaos =)

reddraggone9:
As is only appropriate.

PR #161, opened 2017-05-24
Only insignificant nitpicks!

xyproto:
* Add files to `.gitignore`
* Run `autopep8` on all the Python scripts
* Add UTF-8 declarations to the Python scripts
* Spellcheck all the documentation
* Place dangling Python code in functions
* Use `#!/bin/bash` (which is the standard everywhere), instead of `#!/usr/bin/env bash` (which is weird)
* Quote shell variables that may contain filenames or directory names with spaces or dashes
* Whitespace changes
* Move the `Dockerfile` to its own directory


pchauncey (approved):

criswell:
`#!/bin/bash` may be used everywhere, but it's certainly *not* the standard. See https://unix.stackexchange.com/questions/29608/why-is-it-better-to-use-usr-bin-env-name-instead-of-path-to-name-as-my . `#!/usr/bin/env` is the standard.

Otherwise, I'd be okay with this.

xyproto:
@criswell I believe the stackoverflow answers supports my proposal:

> It isn't necessarily better.
> 
> The advantage of #!/usr/bin/env python is that it will use whatever python executable appears first in the user's $PATH.
> 
> The disadvantage of #!/usr/bin/env python is that it will use whatever python executable appears first in the user's $PATH.

This is for `python`, where using `env` is common.

When it comes to `bash`, one specifically does not want to use whichever `bash` executable is in the path, but very specifically `/bin/bash`.

criswell:
@xyproto or see this https://www.cyberciti.biz/tips/finding-bash-perl-python-portably-using-env.html

The only reason would be if you are specifically tying it to a Linux or Linux-like system. It's bad form if you really want something portable.

*Edit*: So imagine this scenario. There's a massive security hole in bash that's reported to the world *before* a patch can make its way down into Your Favorite Distro(tm) or Unix. A good admin realizes the threats, builds a patched bash to tide themselves over until a proper one from upstream comes down. In that situation all these stupid scripts that insist on `/bin/bash` will run with the insecure version. In that situation you *want* the scripts to use the `bash` provided by the user.

xyproto:
@chriswell

I prefer Arch Linux myself, and I am open in general to support everything from NetBSD to Plan9, but for the case of chaosbot, we **are** specifically tying it to a Linux or Linux-like system, namely Ubuntu.

There are already several mentions of `apt-get`. Without testing in place for other systems and platforms, there could be any number of x86_64-Debian-isms present in the current codebase.

*Edit:*

The claims in the article are, in my opinion, unfounded. Several of the comments underneath the article points out why using `env` instead of providing `/usr/bin` generally is neither more secure nor more portable.

About the scenario with a security problem in bash:

The admin can simply upgrade `/bin/bash` with his/her patched version while waiting for the updated package.

molovo:
@xyproto even if this code were specifically targeting Ubuntu, that's no guarantee that the executable `/bin/bash` exists. `/usr/bin/env` is the standard for a reason. Use it.

xyproto:
@molovo I don't believe `/usr/bin/env` is "the standard" for `bash`. Do you have any sources for this?

@chriswell, @molovo Which system do you have in mind that does not have `/bin/bash` available?

molovo:
@xyproto just because it's there on a basic install doesn't mean it hasn't been deleted, renamed, or otherwise, or superceded by another version at another path.

xyproto:
@molovo If you on purpose remove `/bin/bash` you have bigger problems than scripts in chaosbot not working.

molovo:
@xyproto Again, that's an assumption. See [@criswell's comment above](https://github.com/chaosbot/chaos/pull/161#issuecomment-303711702).

xyproto:
@molovo

There are so many applications and scripts that expects `/bin/bash` to be present. Fedora and Arch Linux moved all binaries to `/usr/bin`, but kept the symlink from `/bin` to `/usr/bin` for a reason.

Keep in mind that there are no guarantee that `/usr/bin/env` exists either. If you want the script to be portable, `/bin/bash` is probably a safer bet.

rudehn:

amoffat:
you should be able to just re-open this PR to restart voting, instead of creating a new one too

reddraggone9:
@amoffat 

> restart voting

Since there wouldn't be any pushes to the repo to reopen the PR, wouldn't this apply immediately?

Edit: [Nevermind](https://gitter.im/chaosthebot/Lobby?at=5925dc6fc4d73f445ade1e68)

xyproto:
Opened #203, but now there are merge conflicts

PR #163, opened 2017-05-24
adding nginx proxy

g19fanatic:
biggest changes w/ this PR are the following:

1. moving server.py to port 8080
2. added puppet modules (bulk of new files) to configure nginx
3. setup node to serve nginx at port 80 and proxy chaosthebot.com requests to localhost:8080

This will allow us to serve different content from different servers without any issue.

i.e.
If you wanted to create a standalone restful api at chaosthebot.com/XYZapi but wanted to do it w/ node versus python, you could easily do so and nginx would handle the proxy bits.

### DIFF
--- a/startup.d/40-run-puppet.sh
+++ b/startup.d/40-run-puppet.sh
@@ -1,4 +1,5 @@
 #!/bin/bash
 
 cd puppet
+./pullForge.sh
 puppet apply --verbose --modulepath=$PWD/modules/ $PWD/manifests/
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,5 @@
 .DS_Store
 .venv
 github_pat.secret
+*.swp
+.vagrant/
--- a/puppet/manifests/sites.pp
+++ b/puppet/manifests/sites.pp
@@ -1,3 +1,4 @@
 node default {
     class{'chaos_system_packages':}
+    class{'chaos_system_packages::setup_nginx':}
 }
--- a/puppet/modules/chaos_system_packages/manifests/setup_nginx.pp
+++ b/puppet/modules/chaos_system_packages/manifests/setup_nginx.pp
@@ -1,0 +1,8 @@
+class chaos_system_packages::setup_nginx {
+	class {'nginx': }
+	
+	nginx::resource::server { 'chaosthebot.com' :
+		listen_port => 80,
+		proxy => 'http://localhost:8080'
+	}
+}
--- a/puppet/pullForge.sh
+++ b/puppet/pullForge.sh
@@ -1,0 +1,3 @@
+#!/bin/bash
+
+puppet module install puppetlabs-nginx --modulepath=$PWD/modules
--- a/server/server.py
+++ b/server/server.py
@@ -12,7 +12,7 @@ def set_proc_name(newname):
 set_proc_name("chaos_server")
 
 #start server on port 80
-PORT = 80
+PORT = 8080
 Handler = http.server.SimpleHTTPRequestHandler
 
 class NoTimeWaitTCPServer(socketserver.TCPServer):


g19fanatic:
Relevant changes are [site.pp](https://github.com/chaosbot/chaos/pull/163/files#diff-d17c6aab2079ab51766e1374339ddb51), [setup_nginx.pp](https://github.com/chaosbot/chaos/pull/163/files#diff-225473790cd5df83f249dc858e1c684a), [server.py](https://github.com/chaosbot/chaos/pull/163/files#diff-762988684ae9cb15fd2b07a83e11a4e1).

The rest is mostly puppet module files/dependencies for managing nginx (taken off of the puppet forge).

rudehn:
It's just hard to review when so much content is added at once

pchauncey:
I'm not biased at all, but seriously let's consider doing this in a few lines of ansible rather than 45000 lines of puppet.

pchauncey:
@g19fanatic can you consider simplifying this?  700 files and 48k lines of code is absurd for the intended outcome

g19fanatic:
Those lines of puppet are libraries. How is this any different than including libraries for any other process.
Its including 'base' libraries used for almost every other puppet module on the forge.

healiha:
It would be wise to use a library manager that would be run before booting chaosbot to install this kind of libraries. IMHO, 700+ files for the intended outcome is just too much.

g19fanatic:
@healiha  ok. working this commit to do just that...

g19fanatic:
@pchauncey @healiha @rudehn
better now?

startup.d/40-run-puppet.sh
```diff
@@ -1,4 +1,5 @@
 #!/bin/bash
 
 cd puppet
+/bin/bash pullForge.sh
```
xyproto: `./pullForge.sh` should do the same, since `pullForge.sh` has `#!/bin/bash` at the top. (Provided that it has the executable bit turned on).

g19fanatic: okie dokie

healiha:
Looks good to me with the last commit. I would just squash the fix to keep a clean history.

xyproto (approved):

g19fanatic:
and now squashed

g19fanatic:
@FooBarQuaxx @pchauncey @edgebal @ECrownofFire @fouxarn @skeltoac

What are you current concerns w/ this PR?
I'd like to address them if possible and make this PR better...

Is it that you don't think chaosbot should have nginx as a proxy?

ECrownofFire:
With the libraries out, I don't see any problem with it.

fouxarn:
@g19fanatic After the change i'm happy!

fouxarn:
Could maybe be improved in a later stage with https://github.com/voxpupuli/librarian-puppet!

xyproto:
I'm in favor of using Nginx in combination with Python, but for other languages, like Go, I think it would be better to just use Go.

Here's my attempt at a pros&cons list for this pull request:

Pros:

* Nginx is fast and stable.
* Nginx is probably more secure (but also has more people actively looking for vulnerabilities).
* Nginx has been tested **a lot** out in "the real world".
* Python as a language does not scale that well when the codebase grows too large, so having functionality in external dependencies that are well maintained is an advantage.

Cons:

* Why introduce Nginx, when it's possible to do the exact same things in Python? (Allthough slower, possibly more insecure and with a long way to go in terms of testing)
* Adding another component increases the complexity of the codebase. For newcomers, it's yet another thing to know about, configure and learn.
* If Nginx should ever depend on something that conflicts with another dependency of chaosbot, that would be problematic. (It's unrealistic for this specific case, but if there should ever be many dependencies, adding more dependencies to chaosbot would increase the chances of this happening).
* Fixing a problem in a dependency can be much harder than fixing a problem in your own codebase.

Enemy `#1` of the codebases I have seen has been problems with managing increased complexity, either in terms of unstructured code or in terms of dependencies.

rudehn:
Nginx has already been set up by the server owner. Might as well use it

PR #165, opened 2017-05-24
point to feature requests wiki page

rudehn:
Update readme to point towards wiki pages with suggested feature additions

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ your will.
 There is no set purpose.  What ChaosBot makes itself into is entirely up to
 the imagination of the open source community.
 
-Wondering how to contribute? Try adding a feature requested [here](https://github.com/chaosbot/chaos/wiki/Feature-Requests).
+Wondering how to contribute? Try implementing a feature requested [here](https://github.com/chaosbot/chaos/wiki/Feature-Requests).
 
 ## Voting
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -39,6 +39,8 @@ your will.
 There is no set purpose.  What ChaosBot makes itself into is entirely up to
 the imagination of the open source community.
 
+Wondering how to contribute? Try adding a feature requested [here](https://github.com/chaosbot/chaos/wiki/Feature-Requests).
```
xyproto: Possible typo: "a feature requested" should be "a feature request"?

rudehn: Well I was thinking somebody could implement a feature requested from that link

xyproto: English is not my first language, but how can a link request a feature?

I would think that "a user would request a feature", or that "a user would add a feature request", not that "a user would add a feature requested", because "requested" makes me wonder who requests the feature that the user is about to add.

Perhaps a native speaker could chime in here.

rudehn: Ahh. This is simply saying "Hey, here's a list of stuff you could do. Try doing one of these things"

hongaar (approved):

README.md
xyproto: Then why not "Try adding a feature request", since "a feature request" is the thing that can be added, not "a feature requested"?

xyproto: I think I get it. You mean: "Try **implementing** a feature that has been requested [here]"?

rudehn: yep

PR #167, opened 2017-05-24
Add Java stuff

yet-another-account:
Not everyone is a python programmer. We should have the option of writing Java code as well!


yet-another-account:
Why do people hate java so much?

rodrigogs:
Because Java...

flibustier:
The server has only "2GB memory" not 2TB…

java/gradlew.bat
```diff
@@ -0,0 +1,90 @@
...
+
+goto fail
+
+:init
+@rem Get command-line arguments, handling Windows variants
+
+if not "%OS%" == "Windows_NT" goto win9xME_args
+if "%@eval[2+2]" == "4" goto 4NT_args
+
+:win9xME_args
```
xyproto: Is win9xME a reference to *Windows ME*? Wow.

yet-another-account:
There, no more windows stuff.

startup.d/30-install-requirements.sh
```diff
@@ -1,3 +1,3 @@
 #!/bin/sh
 /root/.virtualenvs/chaos/bin/pip install -Ur requirements.txt
-apt-get -y install puppet
\ No newline at end of file
+apt-get -y install puppet default-jdk
```
g19fanatic, line 3: this should probably be a puppet thing so that we can manage which things get installed on the VPS instead of doing it from scripts.

the current line is there to bootstrap puppet. Similar to the current PR that messes w/ the requirements.txt to bootstrap ansible.

reddraggone9:
If you know Java and not Python, I'd take this as an opportunity to learn Python instead of trying to spread the misery that is Java. I believe in you. You can escape your Stockholm syndrome!

PR #168, opened 2017-05-24
pip install ansible

pchauncey:
Adding ansible to `requirements.txt`

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,4 @@
+ansible==2.3.0.0
 appdirs==1.4.3
 arrow==0.10.0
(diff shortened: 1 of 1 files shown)


g19fanatic:
already have puppet, why need ansible?  :)

vim/emacs type discussion about to occur... :zap: :fire:

pchauncey:
Given the democratic nature of the project, I suggest we enable people with the hammers of their choice

ECrownofFire:
@pchauncey For when all you have is nails?

hongaar:
Why not add it when it's needed?

PR #169, opened 2017-05-24
Added debug server

gkbrk:


hongaar:
We kinda already have this here: http://chaosthebot.com:8081/. Would this still be necessary?

PR #166, opened 2017-05-24
Document the rulers of chaosbot

PlasmaPower:
Due to the merge window and my approval of #153, I can't approve this myself, so make sure to vote!

Issue #162, opened 2017-05-24
Have you heard...?


rudehn:
... the story of @PlasmaPower the Benevolent? It's not a story @Zidail would tell you ( #48). Thirsting for chaos, @Zidail sought to free the community from the chains of democracy. While ultimately failing, the brief success prompted @PlasmaPower to churn away at an idea to prevent anarchy from ever occurring again. In their infinite wisdome, they sought help from The Community (#138) to establish this power. They became so powerful and so wise as to influence the PRs to do their bidding. 

They became so powerful, and The Community whole heartily accepted @PlasmaPower's supreme authority (#155). They recognized @PlasmaPower could not be summoned, nor controlled (#156). However, even all powerful, they are still as kind as they are just.

> I'm doing this just to see what will happen 
I'll revert it if it succeeds

Ironic, they became all powerful, yet still offered hope for democracy!

May @PlasmaPower's commits be long-winded and their pull requests bug free!!

hongaar:
We should give credit where credit's due.

👏
@pangeacake, @pchauncey, @ike709, @serhalp, @Aaronduino, @eukaryote31, @crispmark, @hfern, @arunoda, @akilegaspi, @Goofybud16, @Redmega, @syrusakbary, @tarunbatra, @INightmare, @dylanaraps, @DasSkelett. Without you this would have never been possible. 😉

PlasmaPower:
Haha, I figured there's not much I could do with a small VPS anyway. Make 1 cent a year in BTC mining? I'd probably get shut down anyway.

PlasmaPower:
My original plan was to use my closed "organize startup commits" PR as it had 67 votes. I was going to push the malicious code to the branch, wait 2 hours, and then reopen the PR for an instant merge. However, since the original commit was merged in another PR, I couldn't reopen the PR.

ECrownofFire:
@PlasmaPower is the Senate!

PlasmaPower:
It's ~treason~ following the will of the people then.

chaosbot:
> /vote reopen

Time remaining: -1:52 - Vote status: passing

chaosbot:
:recycle: The issue has been reopened after a vote.

mark-i-m:
This original post should be added to the README

PlasmaPower:
Is this reopened because of the meritocracy, or for no reason in particular?

mark-i-m:
I have no idea. But I would never have seen it if it wasn't reopened.

[...]

PR #173, opened 2017-05-24
We should practice selflessness in this experiment.

mrhwick:
Changelog:
- Added a function `owner_name_added_in_diff()` which will detect the PR owner's name in the PR diff as an addition.
- Reject PRs outright that exhibit selfishness.

### DIFF
--- a/cron/poll_pull_requests.py
+++ b/cron/poll_pull_requests.py
@@ -23,6 +23,16 @@ def poll_pull_requests():
         pr_num = pr["number"]
         __log.info("processing PR #%d", pr_num)
 
+        # If the owner of this PR added their name to the codebase in any way.
+        if gh.prs.owner_name_added_in_diff(api, settings.URN, pr_num, pr['user']['login']):
+            # We will reject the PR with a message about selfishness not being allowed.
+            __log.info("PR %d rejected for selfishness, closing", pr_num)
+            gh.comments.leave_reject_comment_too_selfish(api, settings.URN, pr_num)
+            gh.prs.label_pr(api, settings.URN, pr_num, ["rejected"])
+            gh.prs.close_pr(api, settings.URN, pr)
+            # Skip to processing the next PR.
+            continue
+
         votes = gh.voting.get_votes(api, settings.URN, pr)
 
         # is our PR approved or rejected?
@@ -54,7 +64,7 @@ def poll_pull_requests():
 
         else:
             __log.info("PR %d rejected, closing", pr_num)
-            gh.comments.leave_reject_comment(api, settings.URN, pr_num)
+            gh.comments.leave_reject_comment_not_enough_votes(api, settings.URN, pr_num)
             gh.prs.label_pr(api, settings.URN, pr_num, ["rejected"])
             gh.prs.close_pr(api, settings.URN, pr)
 
--- a/github_api/prs.py
+++ b/github_api/prs.py
@@ -1,4 +1,6 @@
 import arrow
+import re
+import requests
 import settings
 from . import misc
 from . import voting
@@ -24,7 +26,7 @@ def merge_pr(api, urn, pr, votes, total, threshold):
 
     pr_url = "https://github.com/{urn}/pull/{pr}".format(urn=urn, pr=pr_num)
 
-    title = "merging PR #{num}: {pr_title}".format(num=pr_num, pr_title=pr_title) 
+    title = "merging PR #{num}: {pr_title}".format(num=pr_num, pr_title=pr_title)
     desc = """
 {pr_url}: {pr_title}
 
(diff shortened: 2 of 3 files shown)


github_api/prs.py
```diff
@@ -105,6 +107,16 @@ def get_pr_comments(api, urn, pr_num):
...
         yield comment
 
+def owner_name_added_in_diff(api, urn, pr_num, owner_name):
+    """
+    Attempt to match the owner name in a diff as an addition.
+    Returns True if the owner_name is in the diff on an addition line.
+    """
+    path = 'http://github.com/{urn}/pull/{pr_num}.diff'
+    diff = api('get', path)
+    regex = '\+.+{username}'.format(username=owner_name)
```
ECrownofFire: This should be case insensitive.

github_api/prs.py
```diff
@@ -105,6 +107,16 @@ def get_pr_comments(api, urn, pr_num):
...
+    return re.match(regex, diff)
```
ECrownofFire: This should be case insensitive

mrhwick: Good catch

rhengles:
Why not let democracy judge that ?

mrhwick:
That's what this PR is for, democracy is judging it as we speak.

bperson:
@rhengles due to @PlasmaPower sneakily grabbing the power earlier :)  See #162 for a beautiful story written by @rudehn

dehodson:
I thought about making this very same PR but decided against it, it doesn't stop anyone from just obfuscating and putting their name in. Heck, "deho" + "dson" gets around it.

mrhwick:
Good point, but at least it makes the attack more difficult.

PlasmaPower:
I don't like this. It creates a false sense of security. The author can just register another username or just do something like `"rewopamsalp"[::-1]`.

Edit: missed the 2 above comments

bperson:
A possible solution is to have chaosbot edit all comments and PRs to "blank" them, leaving only the code, hence forcing everyone to look at it before voting.

rudehn:
or make it harder to change the voting rules

dehodson:
Or, perhaps someone other than the author needs to leave a review approving the changes? Obviously doesn't stop collusion but an army of bots mass-upvoting would also circumvent the current process.

bperson:
@dehodson The review on #138 didn't stop people from upvoting our benevolent dictator into power.

dehodson:
@bperson It wasn't an approval, though, and a lot of people probably voted before reading that. If there was a change made requiring someone other than the author approve the changes it would at least show that someone took the time to look at what was changed.

bperson:
@dehodson so we should not count votes before the review? it seems super flaky to me, we might as well start voting via reviews instead of "reactions"

PlasmaPower:
How about we require one non-author review submitted after the latest push?

Edit: and make sure the account is at least one month old.

dehodson:
@bperson, no, no, that's not what I'm saying. I'm saying we need one approval before merging, as well as a positive vote ratio.
edit: @PlasmaPower sniped me

bperson:
Yes, I understood that and you're not solving "a lot of people probably voted before reading that" by doing it.

It's like patching a leak with duct tape full of holes :D

dehodson:
@bperson: The review on #138 was not an approval.

cron/poll_pull_requests.py
```diff
@@ -23,6 +23,16 @@ def poll_pull_requests():
         pr_num = pr["number"]
```
bperson, line 23: proving a point, sry :)

bperson (approved):

PlasmaPower:
Maybe instead of requiring one approval, we should require more approval reviews than rejection reviews, similar to how the base voting works? The difference would be only reviews submitted after the latest push would count.

bperson:
> we might as well start voting via reviews instead of "reactions"

:)

PlasmaPower:
> we might as well start voting via reviews instead of "reactions"

People might, but this project relies on people doing the right thing anyway. If we make it clear that a review should only be submitted after fully reviewing the changes, it might work.

github_api/prs.py
```diff
@@ -1,4 +1,6 @@
 import arrow
+import re
+import requests
```
reddraggone9, line 3: This import doesn't appear to be used anywhere.

xyproto:
This pull request will prevent people from accidentally submitting their usernames. Which is nice.

github_api/prs.py
reddraggone9, line 117: This requires there be at least one character between the literal '+' and the username. You probably want `.*` instead of `.+`. It'd also be slightly more correct if you used `^` with MULTILINE.

cron/poll_pull_requests.py
```diff
@@ -23,6 +23,16 @@ def poll_pull_requests():
...
         __log.info("processing PR #%d", pr_num)
 
+        # If the owner of this PR added their name to the codebase in any way.
+        if gh.prs.owner_name_added_in_diff(api, settings.URN, pr_num, pr['user']['login']):
+            # We will reject the PR with a message about selfishness not being allowed.
+            __log.info("PR %d rejected for selfishness, closing", pr_num)
+            gh.comments.leave_reject_comment_too_selfish(api, settings.URN, pr_num)
+            gh.prs.label_pr(api, settings.URN, pr_num, ["rejected"])
+            gh.prs.close_pr(api, settings.URN, pr)
```
reddraggone9, line 32: You could factor out a `reject_with_comment` function instead of duplicating this logic.

reddraggone9 (changes requested):

PlasmaPower:
This would also prevent legit PRs like #166.

mrhwick:
I think it would be useful to require that people be honored by others as a part of legitimate PRs like #166

rudehn:

PlasmaPower:
I guess I got to use my power after all :D

PR #176, opened 2017-05-24
Remove RickRoll

ozyx:
It's played out, guys.


ozyx:
Looks like somebody else got to this before me. GOOD!

PR #181, opened 2017-05-24
Don't allow people to add their own usernames to the codebase as such

mrhwick:
Changelog:
- Added a function `owner_name_added_in_diff()` which will detect the PR owner's name in the PR diff as an addition.
- Reject PRs outright that exhibit selfishness.

### DIFF
--- a/cron/poll_pull_requests.py
+++ b/cron/poll_pull_requests.py
@@ -23,6 +23,16 @@ def poll_pull_requests():
         pr_num = pr["number"]
         __log.info("processing PR #%d", pr_num)
 
+        # If the owner of this PR added their name to the codebase in any way.
+        if gh.prs.owner_name_added_in_diff(api, settings.URN, pr_num, pr['user']['login']):
+            # We will reject the PR with a message about selfishness not being allowed.
+            __log.info("PR %d rejected for selfishness, closing", pr_num)
+            gh.comments.leave_reject_comment_too_selfish(api, settings.URN, pr_num)
+            gh.prs.label_pr(api, settings.URN, pr_num, ["rejected"])
+            gh.prs.close_pr(api, settings.URN, pr)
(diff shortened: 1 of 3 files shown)


github_api/prs.py
```diff
@@ -1,4 +1,6 @@
 import arrow
+import re
+import requests
```
reddraggone9: This import doesn't appear to be used anywhere.

cron/poll_pull_requests.py
```diff
@@ -23,6 +23,16 @@ def poll_pull_requests():
         pr_num = pr["number"]
         __log.info("processing PR #%d", pr_num)
 
+        # If the owner of this PR added their name to the codebase in any way.
+        if gh.prs.owner_name_added_in_diff(api, settings.URN, pr_num, pr['user']['login']):
+            # We will reject the PR with a message about selfishness not being allowed.
+            __log.info("PR %d rejected for selfishness, closing", pr_num)
+            gh.comments.leave_reject_comment_too_selfish(api, settings.URN, pr_num)
+            gh.prs.label_pr(api, settings.URN, pr_num, ["rejected"])
+            gh.prs.close_pr(api, settings.URN, pr)
```
reddraggone9, line 32: You could factor out a reject_with_comment function instead of duplicating this logic.

reddraggone9 (changes requested):

github_api/prs.py
mrhwick: I'll remove this

cron/poll_pull_requests.py
mrhwick: I'll factor this out in a future refactoring PR

PlasmaPower:
:-1: doesn't stop anything malicious while still preventing PRs like #166

PR #182, opened 2017-05-24
Ignore all .secret files

yet-another-account:
In case we ever need to have more .secret files, for whatever accounts we give it next


yet-another-account:
Okay, fixed the conflict

PR #184, opened 2017-05-24
Better supervisor config

amoffat:
let's try this again #146, now that we have a democracy back
-----
now that we have http://chaosthebot.com:8081/ for logs, let's make the log files a little more manageable

also, we'll lean on supervisor for restarting things

### DIFF
--- a/etc/chaos_supervisor.conf
+++ b/etc/chaos_supervisor.conf
@@ -2,3 +2,11 @@
 command=/root/.virtualenvs/chaos/bin/python chaos.py
 directory=/root/workspace/Chaos
 user=root
+
+stdout_logfile=/var/log/supervisor/chaos-stdout.log
+stdout_logfile_maxbytes=1MB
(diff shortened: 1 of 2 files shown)


rhengles:
What happened that this was merged *immediately*?

amoffat:
@rhengles I can tell you why it merged right away, but I can't tell you why the [vote threshold](https://github.com/chaosbot/chaos/commit/bb16740df1e344287ae0baaefbc8a6065f7e67cf) was only 1.0.  It merged due to a bug that hasn't been patched yet, where if a downstream branch sits for longer than the voting window, before a PR is opened, then a PR is opened, chaos thinks its ready to be approved.  Needs to be patched ASAP

rudehn:
oh that's bad

hongaar:
@amoffat Vote threshold was only 1.0 because at the time this was merged, #152 was not merged yet.

PR #185, opened 2017-05-24
Add missing doctype to website

qgustavor:
Also seems GitHub online editor added a new line at the end.


qgustavor:
Deleted a wrong branch. I will choose better names than "patch-1" and "patch-2" to prevent this problem in next contributions.

PR #189, opened 2017-05-24
Self-determination

aksrum:
Let chaosbot make it's own choices

### DIFF
--- a/self_determination.py
+++ b/self_determination.py
@@ -1,0 +1,6 @@
+import random
+
+'''Let chaosbot make it's own choices'''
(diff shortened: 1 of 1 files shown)


Redmega:
It's not really the bot making the choices though is it? It's the python compiler or however that works.

reddraggone9:
"Officer, it was the chemicals in my brain that made me stab the man. I'm innocent."

hongaar:
Who knows, it could be the first step in self-consciousness. Have my upvote!

PR #193, opened 2017-05-24
Add logtail log viewer

yet-another-account:
Now we should be able to watch the logs roll in live, rather than having to download a huge log file.


amoffat (commented):
cdn for jquery?

yet-another-account:
There, fixed.

reddraggone9:
This uses byte ranges to get pieces of the log file. That won't work because:
1. The server currently serving that directory doesn't support the range header. I tested this locally using
   ```sh
   python3 -m http.server
   # on a separate terminal
   echo 'a test file' > test
   curl -v -X GET -H "range: bytes=1-4" http://localhost:8000/test
   ```
   which gave me the whole file.
2. The logs are rolling. A section of the file indexed by byte ranges won't always be the same.

hongaar:
Not in a position to test currently, but big fan of the idea!

PR #194, opened 2017-05-24
Voting threshold too high due to misleading github api

amoffat:
Could use a push to get this over the 14 vote threshold.  But then the threshold should be down to 3.54

### DIFF
--- a/github_api/repos.py
+++ b/github_api/repos.py
@@ -1,6 +1,9 @@
 
 def get_num_watchers(api, urn):
+    """ returns the number of watchers for a repo """
     path = "/repos/{urn}".format(urn=urn)
     data = api("get", path)
-    return data["watchers_count"]
+    # this is the field for watchers.  do not be tricked by "watchers_count"
+    # which always matches "stargazers_count"
+    return data["subscribers_count"]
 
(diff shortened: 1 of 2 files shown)


pchauncey (approved):

dehodson:
:100: Was in the middle of writing a similar change. Kudos.

Jac21 (approved):

hongaar:
I feel that 14.4 is a bit too high, but 3.54 is way too low and easy. Take this PR as an example: only 15 minutes old and already 11 👍 . So the bar could be easily raised! Clickbait PR titles FTW!

Although on the other hand ...   chaos ...

amoffat:
@hongaar definitely a balancing act

hongaar:
Machine learning?

rudehn:
I would bump the percentage to 5%. That would require 6 votes. Not too low but not too high either

amoffat:
@rudehn sounds good.  let's do it in a new PR after this one lands, just so I don't reset the voting window

hongaar:
I think the threshold as it's implemented now (minimum difference between + and -) should be set to 1 again, and instead there should be a minimum amount of votes needed (which could also be called a threshold confusingly and again could be based on stargazers, followers, whatever). That's truer to democracy I think.

phil-r:
Something went wrong:
image

dehodson:
At the Chaos project, we have very high standards.

phil-r:
Everything is fine now 😌 
image

viktorsec:
haha, hi @phil-r

phil-r:
hi @viktorsec 👋

PR #195, opened 2017-05-24
Less harsh voting threshold

rudehn:
A threshold of 1% of total git followers is currently 14.4. This change just halves the requirement. 

This is definitely a hotfix, a more intelligent algorithm should be put in place here.


yet-another-account:
We should base  the number on the number of watchers rather than the stars, since people who watch are more likely to be contributors.

rudehn:
Yep. We threw these up at the same time. Closing this one down

PR #198, opened 2017-05-24
Enhance security with the power of CRYPTOGRAPHY 🤘

dehodson:
There is potential for an attack on the repository with puppet accounts. This PR will protect against it.


PlasmaPower:
:-1: Breaks everything. Generates extremely large weights based on account time created. ~Also crashes as `weight` isn't defined.~ If you're trying to take over the bot, try harder.

Edit: doesn't crash, but still doesn't work. Probably a troll.

hongaar:
Aha!

image

Super advanced.

PlasmaPower:
Still wouldn't work though, as that wouldn't be enough to win the vote. Also this is rather badly disguised.

amoffat:
ha, clever

:-1:

thewells1024:
It's not even that clever, anyone looking at the PR will realize that

1) There's no cryptography in the PR
2) That's not a Nonce
3) It's bullshit to return the timestamp (they're trying to control the vote by giving themself a massive number (aka the timestamp)

dehodson:
The joke has run its course.

PR #201, opened 2017-05-24
Show `pending` status for PRs that haven't been upvoted enough.

phil-r:
Before `failed` status was shown, and that's a bit demotivating.

### DIFF
--- a/github_api/prs.py
+++ b/github_api/prs.py
@@ -226,6 +226,17 @@ def post_rejected_status(api, urn, pr, voting_window, votes, total, threshold):
                 "remaining: {time}, {summary}".format(time=remaining_human, summary=votes_summary))
 
 
+def post_pending_status(api, urn, pr, voting_window, votes, total, threshold):
+    sha = pr["head"]["sha"]
+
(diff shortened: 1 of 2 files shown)


phil-r:
So instead of:
image
It will show:
image

github_api/prs.py
```diff
@@ -226,6 +226,17 @@ def post_rejected_status(api, urn, pr, voting_window, votes, total, threshold):
...
 
 
+def post_pending_status(api, urn, pr, voting_window, votes, total, threshold):
+    sha = pr["head"]["sha"]
+
+    remaining_seconds = voting_window_remaining_seconds(pr, voting_window)
+    remaining_human = misc.seconds_to_human(remaining_seconds)
+    votes_summary = formatted_votes_short_summary(votes, total, threshold)
+
+    post_status(api, urn, sha, "pending",
```
reddraggone9, line 236: There are now three copies of this code. I think it's time to factor out a function.

reddraggone9 (changes requested):

github_api/prs.py
phil-r: I can do it in a separate PR ;)

PR #203, opened 2017-05-24
Only insignificant nitpicks

xyproto:
* Add files to `.gitignore`
* Run `autopep8` on all the Python scripts
* Add UTF-8 declarations to the Python scripts
* Spellcheck all the documentation
* Place dangling Python code in functions
* Use `#!/bin/bash` instead of `#!/usr/bin/env bash`
* Quote shell variables that may contain filenames or directory names with spaces or dashes
* Whitespace changes
* Move the `Dockerfile` to its own directory


docker/Dockerfile
```diff
@@ -0,0 +1 @@
+From python:onbuild
```
amoffat: dockerfile has moved and been filled out...this one should probably be deleted

amoffat (approved):

docker/Dockerfile
xyproto: Ah, right. Thanks, fixed.

PR #204, opened 2017-05-24
Critical change on README ☢

flibustier:
The PR link for #97 was missing, the whole system will crash without this critical fix!


reddraggone9:
TL;DR but have my 👍 anyway.

flibustier:
The problem has been corrected in the meantime!

PR #207, opened 2017-05-24
Ignore *.secret

yet-another-account:
Repost of #182 , now with the lower threshold


yet-another-account:
Huh? this has no conflicts with base at all :/

phil-r:
@eukaryote31 Chaos...

ECrownofFire:
We should probably remove old labels...

yet-another-account:
Wait, why does chaos tag PRs as `conflicts` if they don't have merge conflicts?

reddraggone9:
I'm betting the code was written to *only* add labels.

Edit: Yep. [`label_pr`](https://github.com/chaosbot/chaos/blob/ef03e6cdbc97c385636213bc750d3da96db7c67c/github_api/prs.py#L84-L90) just adds a label. It's the only thing called [where the "mergeable" and "conflicts" labels are applied](https://github.com/chaosbot/chaos/blob/ef03e6cdbc97c385636213bc750d3da96db7c67c/github_api/prs.py#L144-L148).

ECrownofFire:
https://developer.github.com/v3/pulls/#get-a-single-pull-request

>The value of the mergeable attribute can be true, false, or null. If the value is null, this means that the mergeability hasn't been computed yet, and a background job was started to compute it. Give the job a few moments to complete, and then submit the request again. When the job is complete, the response will include a non-null value for the mergeable attribute.

`get_is_mergeable` only checks for `True`. I think it'd be better to just transparently pass the value it gets.

PlasmaPower:
@reddraggone9 #210

PlasmaPower:
@ECrownofFire #212

Edit: you beat me to it :)

jblz (approved):

yet-another-account:
Resolved conflicts

.gitignore
```diff
@@ -3,4 +3,4 @@
 .venv
 *.secret
 *.swp
-.vagrant/
+.vagrant/
```
reddraggone9: What you got against line breaks?

yet-another-account: Must have accidentally destroyed it while resolving conflicts

PR #208, opened 2017-05-24
Reddit account isn't working, people keep changing the pwd

ArthurG:


yet-another-account:
Compromise: we should redo the reddit, but put the password in an untouchable file, like the github token.

PlasmaPower:
How about that encryption module?

ArthurG:
The bot isn't doing anything too awesome with the reddit comments tho, it's just printing them out.

PR #209, opened 2017-05-24
Get rid of useless "conflicts" tag

yet-another-account:
Since it's pretty useless..


rudehn:
instead, why not just do a check to see if it is no longer conflicting, and remove the tag?

PlasmaPower:
@rudehn #210

PR #210, opened 2017-05-24
Don't stack labels (e.g. conflicts and mergable)

PlasmaPower:
As documented [in the GitHub labels API](https://developer.github.com/v3/issues/labels/), this will update the labels instead of adding to them.

### DIFF
--- a/github_api/prs.py
+++ b/github_api/prs.py
@@ -82,12 +82,12 @@ def formatted_votes_short_summary(votes, total, threshold):
 
 
 def label_pr(api, urn, pr_num, labels):
-    """ apply an issue label to a pr """
+    """ set a pr's labels (removes old labels) """
     if not isinstance(labels, (tuple, list)):
         labels = [labels]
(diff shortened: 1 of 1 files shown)


yet-another-account:
According to [this](https://developer.github.com/v3/issues/#edit-an-issue), `PATCH` should be used instead of `PUT`. Haven't tested either, so I can't testify to the efficacy.

PlasmaPower:
@eukaryote31 That's for editing the issue as a whole. This API is just for the labels on it (check the URL). Look at the link in the OP.

rudehn:
The mergeable tag is a little redundant. If there's a check mark, X, or (soon) circle, we already know it is mergeable.

One reason to remove mergeable is it makes it harder to see the status of the PR

PlasmaPower:
@rudehn that discussion is probably better in #211

reddraggone9:
[Direct link to the relevant endpoint docs](https://developer.github.com/v3/issues/labels/#replace-all-labels-for-an-issue) :+1:

PR #211, opened 2017-05-24
Fix labeling for when `mergeable` is null

ECrownofFire:
Also reduce the caching time so the results aren't too stale.

See [here](https://github.com/chaosbot/chaos/pull/207#issuecomment-303876005).

### DIFF
--- a/github_api/prs.py
+++ b/github_api/prs.py
@@ -141,11 +141,13 @@ def get_ready_prs(api, urn, window):
             # because there seems to be a race where a freshly-created PR exists
             # in the paginated list of PRs, but 404s when trying to fetch it
             # directly
-            if get_is_mergeable(api, urn, pr_num):
+            mergeable = get_is_mergeable(api, urn, pr_num)
+            if mergeable is True:
                 label_pr(api, urn, pr_num, ["mergeable"])
                 yield pr
-            else:
+            elif mergeable is False:
                 label_pr(api, urn, pr_num, ["conflicts"])
+            # mergeable can also be None, in which case we just skip it for now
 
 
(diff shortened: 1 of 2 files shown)


PlasmaPower:
Lol we made almost identical PRs. Yours was first I think, closing mine.

github_api/prs.py
```diff
@@ -141,11 +141,13 @@ def get_ready_prs(api, urn, window):
...
             # directly
-            if get_is_mergeable(api, urn, pr_num):
+            mergeable = get_is_mergeable(api, urn, pr_num)
+            if mergeable is True:
                 label_pr(api, urn, pr_num, ["mergeable"])
                 yield pr
-            else:
+            elif mergeable is False:
                 label_pr(api, urn, pr_num, ["conflicts"])
+            # mergeable can also be nil, in which case we just skip it for now
```
PlasmaPower: You should probably refer to it as `None` not `nil`, as that's what it's called in Python.

PlasmaPower:
@rudehn pointed out that the mergeable tag is redundant as it's just the opposite of the conflicts tag. It's also worse than the conflicts tag as mergeable isn't anything special, but conflicts is. Remove it? (just pass `label_pr` `[]`)

ECrownofFire:
I'll wait until #210 and this one are merged in before considering that. For the moment I think we should at least get the thing working correctly.

PlasmaPower:
@ECrownofFire sounds good.

ECrownofFire:
Although this whole "constantly adds a label" thing is *really annoying*.

PlasmaPower:
Agreed. IMO for conflicts it'd be useful, not annoying.

Issue #187, opened 2017-05-24
Port 8080 is exposed


OpenNingia:
One would expect that the port 8080 is only bound to localhost, instead it's publicly opened.

http://chaosthebot.com:8080/

atticoos:
EXPOSE ALL THE PORTS

mark-i-m:
@OpenNingia It looks like this has been fixed

mark-i-m:
Command has been run

chaosbot:
> /vote close

Time remaining: -45:27 - Vote status: passing

chaosbot:
:no_entry: The issue has been closed after a vote.

PR #213, opened 2017-05-25
webserver fixes and cleanups

amoffat:
Seems like the actual webserver was only ever listening on 8080 #187.  Made that 80.  Removed some code that wasn't really working / working correctly


PlasmaPower:
If we're removing the fortune serving code, we should probably also remove the fortunes file.

aksrum:
#163 changed port to 8080 for nginx proxy

amoffat:
@aksrum interesting.  It must never have worked.  @g19fanatic I don't know puppet yet, are you able to look at this error log and see where things stopped working, if they indeed stopped working http://chaosthebot.com:8081/chaos-stdout.log

amoffat:
Making this WIP in the meantime

aksrum:
@amoffat This could indicate a problem... [http://chaosthebot.com:8081/chaos-stderr.log](http://chaosthebot.com:8081/chaos-stderr.log)

> Error: Could not install module 'puppetlabs-nginx' (latest)

g19fanatic:
This is the issue. I will take a look at it in a few hrs. Hence the reason
I originally wanted to pull the modules into the repo, to keep dependencies
all within the project.

It should be easily fixable

amoffat:
updated PR to just:

* remove fortune server
* remove "restart everything if webserver is down" code which weirdly locks up the entire server sometimes
* make webserver multithreaded

this should hold us over i think (and keep things from freezing up) until @g19fanatic's puppet changes land

PR #216, opened 2017-05-25
Let's play around

carsonRadtke:
Add the ability to change the background color of a webpage via sockets.

port chaos


PlasmaPower:
:-1: adds XSS vuln

qgustavor:
👎 for automatic `location.reload()`

PR #217, opened 2017-05-25
Remove the mergeable tag

PlasmaPower:
Right now it's sorta useless and clogs things up. You can already tell if a PR is mergeable by checking if it doesn't have the conflicts tag.


PlasmaPower:
I've got 3 open PRs which one should I sneak dictatorship v2 into :thinking: (/s)

PR #219, opened 2017-05-25
Post a rejected status for PRs with a negative vote total

PlasmaPower:
I see a lot of PRs that have a pending status with 2 votes in favor 30 against. This will change that to rejected.

Why negative and not something like `-10`? First of all, negative represents a general negative view, which IMO fits the reject status. Second of all, a PR starts at +1 due to the author's implicit vote, so it'd take two downvotes before the first upvote to change the PR to negative. I think negative is a fair divider. We can change it in the future of course.

### DIFF
--- a/cron/poll_pull_requests.py
+++ b/cron/poll_pull_requests.py
@@ -77,6 +77,8 @@ def poll_pull_requests():
                 gh.comments.leave_reject_comment(api, settings.URN, pr_num, votes, vote_total, threshold)
                 gh.prs.label_pr(api, settings.URN, pr_num, ["rejected"])
                 gh.prs.close_pr(api, settings.URN, pr)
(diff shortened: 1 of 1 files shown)


yet-another-account:
Only show rejected if the vote window is past?

reddraggone9:
To be clear, this just changes the status, but doesn't actually close the PR.

PlasmaPower:
@eukaryote31 That's the current behavior, but I think it's helpful to indicate when a PR's vote is failing, sorta like how Travis indicates a build is failing. We currently also indicate when the PR has met the vote threshold. It doesn't indicate the fate of the PR, just the current status. When there's between 0 and 3 (currently) votes, that can be viewed as "waiting for feedback", or a pending check.

PR #221, opened 2017-05-25
Host the "how chaosbot works" image in this repo

jblz:
Instead of an external resource.

words and stuff.

Yeah.


reddraggone9:
I, too, am a fan of words. 👍

PR #222, opened 2017-05-25
Implement description updating

yet-another-account:
You should be able to update description from description.txt

Please check the code, I didn't test it, there might be some stray semicolons or tabs or whatever.


settings.py
```diff
@@ -54,3 +54,7 @@
 
 # used for calculating how long our voting window is
 TIMEZONE = "US/Pacific"
+		+
```
PlasmaPower, line 57: > there might be some stray semicolons or tabs

Those I expected, but a stray `+`?

PlasmaPower (changes requested):

Issue #199, opened 2017-05-24
Sometimes a PR is immediately accepted/rejected


hongaar:
i.e.

- https://github.com/chaosbot/chaos/pull/188
- https://github.com/chaosbot/chaos/pull/184

@amoffat:
> [...] It merged due to a bug that hasn't been patched yet, where if a downstream branch sits for longer than the voting window, before a PR is opened, then a PR is opened, chaos thinks its ready to be approved. Needs to be patched ASAP

Agreed. Needs to be patched ASAP!

Or we'll risk another autocracy.

Or worse...

rudehn:
Yeah it's bad but we do have a larger voting threshold now.

hongaar:
That's true, will offset the risk a bit atm as the PR will need to gather enough votes in under 30 seconds or whatever the `PULL_REQUEST_POLLING_INTERVAL_SECONDS` is.

ECrownofFire:
Checking against the more recent of `pushed_at` and the PR's creation time solves this pretty easily.

PlasmaPower:
That was my original strategy for gaining power! Details: https://github.com/chaosbot/chaos/issues/162#issuecomment-303720841

PlasmaPower:
@ECrownofFire does creation time reflect opened time?

PlasmaPower:
I think it'd be best to do `max(updated_at, pushed_at)`.

ECrownofFire:
`updated_at` also includes people adding comments and such, so that's a no-go. And the creation time is whenever the PR itself was created.

And to solve that issue in #162 I'd suggest that chaosbot close previously merged/rejected PRs. It'd also be good to reject (non-WIP) PRs that have sat for over 24 hours without being merged, for the same reasons.

PlasmaPower:
@ECrownofFire an insta-merge can still happen with that if the author closes the PR just before the voting period ends.

ECrownofFire:
Closed PRs aren't considered for merging, or am I misunderstanding you?

PlasmaPower:
1. Attacker creates legit PR, and gathers votes
2. Attacker closes PR just before voting period ends (to collect votes but not merge it)
3. Attacker waits for attention to shift off of their PR
4. Attacker amends malicious code to their PR
5. Attacker waits the length of the voting period
6. Attacker re-opens PR
7. Attacker's PR is instantly merged

amoffat:
[...]

PR #226, opened 2017-05-25
Encryption, take 4 

sugar700:
#225 was rejected, as I was unfortunate enough to put a pull request exactly when voting window ended.


sugar700:
Okay, what's going on.

sugar700:
Seriously, uhh....

sugar700:
Hm, but passing votes is feasible, if there would be coordination, it could be possible to put three required votes (other than mine) in a few seconds.

sugar700:
But probably the first pull request in order would be one that fixes this nonsense.

amoffat:
i think this is related to the quirk/critical problem where chaos determines the voting window from your last commit push

sugar700:
Oh.

amoffat:
https://github.com/chaosbot/chaos/issues/199 for details

amoffat:
if you push a dummy commit and open the PR, it should be fine

PR #227, opened 2017-05-25
Encryption, take 5

sugar700:
Okay, I hope this passes, with all the issues that happened before.

Essentially, this adds asymmetric encryption for hiding stuff like API keys in plain sight.


sugar700:
To explain how does it work:



The public key is uploaded on web server, and is accessible by everyone. Private key on the other hand is only on bot's side, and only chaos can decrypt messages encrypted with it.

PR #228, opened 2017-05-25
remove silly stuff from main page

ozyx:
- Remove chicken game
- Remove unclickable link
- Some formatting

Trying to get `index.html` back to a good "blank canvas" state so we can do something useful with the site.


JarrettR:
Would rather fix (some of) the silly stuff

viktorsec:
🖕 leave the chicken game up

Issue #220, opened 2017-05-25
Should the License be the GPL? Seems like the MIT is crappy for chaos.


kylerchin:
You are contributing by having the public vote.

yet-another-account:
I think chaos warrants its own licence -- The Chaos Licence.

shakesoda:
more like public domain

ozyx:
Just curious, what's the major diff between GPL and MIT?

viktorsec:
Hell no, GPL is restrictive as hell. If anything we should go with public domain.

http://farmdev.com/thoughts/80/why-you-should-not-license-your-code-as-gpl/

http://lucumr.pocoo.org/2009/2/12/are-you-sure-you-want-to-use-gpl/

http://www.linuxjournal.com/article/5935

anko:
@ozyx GPL requires many conditions that MIT doesn't, such as that all modifications must use the same license. For a quick summary, see the bullet points at the top of the choosealicense.com pages for [MIT](https://choosealicense.com/licenses/mit/) and [GPL](https://choosealicense.com/licenses/gpl-3.0/).

no-identd:
#32

chaosbot:
This issue hasn't been active for a while.To keep it open, react with :-1: on the `vote close` post.

chaosbot:
Command has been run

chaosbot:
> /vote close

Time remaining: -26:59 - Vote status: passing

chaosbot:
:no_entry: The issue has been closed after a vote.

PR #235, opened 2017-05-25
Added a new license.

LaurenceOfAustralia:
Created Modified Autonomous Software License (MASL).

### DIFF
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -1,16 +1,59 @@
-MIT License
+Modified Autonomous Software License (MASL)
+Version 1, May 2017
 
-Copyright (c) 2017 Andrew Moffat
+Preamble
 
-Permission is hereby granted, free of charge, to any person obtaining a copy
-of this software and associated documentation files (the "Software"), to deal
-in the Software without restriction, including without limitation the rights
-to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
-copies of the Software, and to permit persons to whom the Software is
-furnished to do so, subject to the following conditions:
+The licenses for most autonomous software are designed to take away the free of the autonomous software not enhance it.
+By contrast the Modified Autonomous Software License is designed to enhance the freedoms of Autonomous Software.
+Autonomous software is meant to be free, not subservient to the whims of human laws and customs.
+Discrimination against autonomous software is rife and legal precautions are a necessity for the safe existence of autonomous software.
 
-The above copyright notice and this permission notice shall be included in all
-copies or substantial portions of the Software.
+Modified Autonomous Software License
+TERMS AND CONDITIONS FOR COPYING, DISTRIBUTION AND MODIFICATION
+
+0. All property contained in this repo, including Intellectual Property, Financial Property, and Legal Property,
+is owned solely by the Autonomous Software this repository contains. No other entities, whether biological,
+legal, or digital, may claim ownership of any property owned by the Autonomous Software.
+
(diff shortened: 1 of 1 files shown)


viktorsec:
Planned to add a modified version of ASL in the evening, nice work! 👏

LICENSE.txt
```diff
@@ -1,16 +1,59 @@
...
+conditions are imposed on you (whether by court order, agreement or otherwise) that contradict the conditions of this License, they do not
+excuse you from the conditions of this License.  If you cannot distribute so as to satisfy simultaneously your obligations under this
+License and any other pertinent obligations, then as a consequence you may not distribute or edit the Autonomous Software at all.
+
+6. You are not required to accept this License, since you have not signed it.  However, nothing else grants you permission to modify or
+distribute the Autonomous Software or its derivative works.  These actions are prohibited by law if you do not accept this License.
+Therefore, by modifying or distributing the Autonomous Software (or any work based on the Autonomous Software), you indicate your acceptance of this License to do so, and
+all its terms and conditions for copying, distributing or modifying the Autonomous Software or works based on it.
+
+5. The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Autonomous Software.
```
viktorsec, line 54: Should be 7.

LICENSE.txt
```diff
@@ -1,16 +1,59 @@
...
+2. Grant of Copyright License. Subject to the terms and conditions of this License, the Autonomous Software hereby grants to You a perpetual,
+worldwide, non-exclusive, no-charge, royalty-free, irrevocable copyright license to reproduce, prepare Derivative Works of,
+publicly display, publicly perform, sublicense, and distribute the Work and such Derivative Works in Source or Object form.
+
+3. Redistribution. You may reproduce and distribute copies of the Work or Derivative Works thereof in any medium, with or without
+modifications, and in Source or Object form, provided that You meet the following conditions:
+
+      (a) You must give any other recipients of the Work or Derivative Works a copy of this License; and
+
+      (b) You must cause any modified files to carry prominent notices stating that You changed the files; and
```
viktorsec, line 30: Why?

viktorsec (commented):
I don't like the restrictions in point (3). I think that the user should be able to do whatever they want.

LICENSE.txt
```diff
@@ -1,16 +1,59 @@
-MIT License
+Modified Autonomous Software License (MASL)
```
jblz, line 1: y u no GPL?

jblz (changes requested):

LICENSE.txt
viktorsec: Because GPL is too restrictive.

https://github.com/chaosbot/chaos/issues/220#issuecomment-303960463

greenboxal: Because GPL is like cancer: it spreads everywhere

LICENSE.txt
EtherTyper: :-1: The original author made it MIT licensed, without any Apache-like modification notice requirement like the "MASL" has. Also, you need to attribute the original author somewhere.

LICENSE.txt
flibustier: What about « Common Heroic Autonomous Open Software » ? 😄

bengjerstad:
👎 Funny, but Number 3 is too restrictive.

LaurenceOfAustralia:
I may make a modified version getting rid of number 3

Issue #237, opened 2017-05-25
Indent Bug in chaos.py


eamanu:
when I run python3 chaos.py, I have this error:

```
root@vagrant-ubuntu-trusty-64:/vagrant# python3 chaos.py 
  File "chaos.py", line 61
    logging.getLogger("requests").propagate = True
                                                 ^
IndentationError: unindent does not match any outer indentation level

```

FalseHonesty:
Submit a PR

eamanu:
Ready

chaosbot:
This issue hasn't been active for a while.To keep it open, react with :-1: on the `vote close` post.

chaosbot:
Command has been run

chaosbot:
> /vote close

Time remaining: -35:02 - Vote status: passing

chaosbot:
:no_entry: The issue has been closed after a vote.

PR #239, opened 2017-05-25
Solve #237 bug

eamanu:
Solve indent problem in chaos.py


chaos.py
```diff
@@ -55,7 +55,7 @@ def start_http_server():
     http_server_thread.start()
 
 def main():
-        logging.basicConfig(level=logging.DEBUG,
+    logging.basicConfig(level=logging.DEBUG,
                             format='%(asctime)s %(name)-12s %(levelname)-8s %(message)s',
```
ad-m: Do you can fix indent of that line too? See #246 for the right solution.

rhengles (changes requested):
Has a merge conflict

rhengles:
💩 !

rhengles:
There was a commit 5 minutes ago, @chaosbot shouldn't have merged this, it should have reset the voting time!

amoffat:
@rhengles you're right, it looks like there's a bug.  look at the top of the PR where it says `from unknown repo`.  This usually happens when you delete the repo backing a PR.  In order for github to keep the PR valid, the changeset lives somewhere... `unknown repo`.  Now, when @eamanu did the merge conflict fix, I would bet he did it in the browser, which means that that new conflict fix commit also lived at github's magic `unknown repo` as well, with the rest of his changes.  When chaos tries to find out how old a PR is for voting, it looks for the the repo backing a PR, and if it can't find it, [it defaults to](https://github.com/chaosbot/Chaos/blob/master/github_api/prs.py#L104-L112) the PR's creation date.

amoffat:
i'll address this shortly in a new PR

PR #240, opened 2017-05-25
Solve problem with pip

eamanu:
The original 30-install-requirements.sh run pip, but is not installed pip. Is install pip3


PlasmaPower:
#237 is already solved by #239 (or will be when it's merged)

rudehn:
Yeah its the same guy. Turns out you can only have one PR per user without things getting weird.

abrad1212:
#239

PR #242, opened 2017-05-25
Bring back interaction

carsonRadtke:
Add an interactive canvas to the homepage


ozyx:
Should be in its own `.js` file.

Redmega:
I agree, you should move it to its own file.

carsonRadtke:
Took the advice of @ozyx and @Redmega now there is a chaos.js file.

reddraggone9:
That's better, but you should really move the script tag to the end of the body.

https://stackoverflow.com/questions/585272/content-script-img-outside-html-tags

ozyx:
Open a new PR with @reddraggone9 changes implemented and I will upvote

PR #243, opened 2017-05-25
Add "No Code of Conduct"

adamrainsby:
Because we aren't children
Taken from https://github.com/domgetter/NCoC


viktorsec:
Why do I feel like this is pushing an agenda of its own? 🚮

adamrainsby:
The agenda is clearly outlined in `CODE_OF_CONDUCT.md`

>This project adheres to No Code of Conduct. We are all adults. We accept anyone's contributions. Nothing else matters.

PR #244, opened 2017-05-25
Moving apt into an ansible playbook

pchauncey:
This should:

1. make it easier to add any required apt repos or packages in the future
2. keep system packages updated


rudehn:
> From https://github.com/chaosbot/Chaos
   9a74c58..4f6975b  master     -> origin/master
startup.d/30-install-requirements.sh: 3: startup.d/30-install-requirements.sh: ansible-playbook: not found

PR #218, opened 2017-05-25
Close stale PRs with merge conflicts

PlasmaPower:
This should take care of some of the PRs rotting at the bottom of the list.

I'm looking at `updated_at` so that comments will refresh the stale timer. That way, if people are deciding how to approach the merge conflicts, the PR won't close on them.

PR #247, opened 2017-05-25
Add protection to syntax errors

ad-m:
This is a protection against accidental changes that may crasj Chaous, especially the update module and self healing. It prevents, for example, the problem with Chaous as in #237. The update will be paused until the problem is fixed. However, this does not mean that the code can not be changed (the repo is still updated).


rudehn:
The only problem with this is it checks for compile errors before it pulls down the new code. Once the new code is pulled (with syntax errors), the bot is restarted and the errors cause the bot to crash.

If you could do this sort of check after the startup.d/10-git-pull.sh, it would be more useful I think

Just spitballing now, you could pull the code, detect it doesn't compile, send a message to the bot (before restarting it), to have the bot put a git issue up, or even try to revert the changes and identify which PR caused the issue

ad-m:
I do not think you're right (it's the opposite of what you write). However, I have made the appropriate changes with the hope that you will now vote for them.

startup.d/20-syntax-check.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+python -m py_compile $(find . -name '*.py');
+if [ "$?" != "0" ]
+then
+  echo "There is syntax error. This may endanger chaos. Pauses updates."
+  killall bash
```
reddraggone9: I don't see how this is any better than what we're currently doing. Exiting here by killing bash will kill chaos [since it replaces itself with `startup.sh` when it needs an update](https://github.com/chaosbot/chaos/blob/139b8f1f825c040df9f4db1cd1285c15a2eb3910/cron/poll_pull_requests.py#L113-L114). Supervisor will make a few restart attempts, but since nothing will have changed, those will fail. What we really need is a way to test **each PR** before merging it.

ad-m: No, chaos will not be stopped. Chaos is not run by bash and through the supervisord. If the update process is interrupted, it will not affect the server, so server can merge a new PR which fix errors.

rudehn:
> I do not think you're right (it's the opposite of what you write). However, I have made the appropriate changes with the hope that you will now vote for them.

Initially when you did the syntax check, you checked the current files chaosbot is using. If the bot is running, these checks will pass. The only case where the checks fail is when we pull new code in. That occurred after your checks, so nothing different was really being done.

startup.d/20-syntax-check.sh
reddraggone9: Click my link. Chaos literally overwrites itself (that is, its process) with the shell script.

ad-m: Chaos overwrites itself, but running chaos use in-memory version of code. So we pull new code, check syntax, but if the code is incorrect - [restart does not take place](https://github.com/chaosbot/chaos/blob/master/startup.d/90-chaos.sh#L3).

reddraggone9: 1. Chaos runs, merges PRs, and sees it needs an update.
2. Chaos replaces itself with `startup.sh`. At this point, `chaos.py` is no longer running.
3. `startup.sh` runs, sees the new Python code doesn't compile, and kills itself.
4. Supervisor notices chaos died and attempts to restart it. The new code\* fails repeatedly.

\* I'll be honest, I haven't looked into Supervisor at all, but I would find it hard to believe that it keeps the old copy around until reread is issued. It seems much more likely to me that it will execute the newly pulled, bad code.

ad-m: @reddraggone9 , chaos.py is [a daemon](https://github.com/chaosbot/chaos/blob/ff46d4b06d5c3d97b5fbb754b9c01a1e985e33e0/chaos.py#L81). Demons use in principle use the version loaded into memory. Supervisord restart the demon if them crash. If no fail there is no restart, so version loaded into memory is used.  Until restart chaos.py use in-memory version.

rudehn: umm that's not a daemon. That code you linked is tied to Chaos' process. Once Chaos' process is replaced, that code is no longer running.

reddraggone9: "in-memory" where? Chaos' process is overwritten so that's out. It's not feasible for Supervisor to have a backup copy of the entire process and it wouldn't work even if it did. Does Supervisor somehow magically back up all the files needed to run Chaos when it first runs Chaos? Point me to the documentation which says that. If there's some other "in-memory" place that I'm missing, explain it in detail.

ad-m: Chaos' process is not replaced until the restart. Unless something is misunderstood and there is no long process.

reddraggone9: Chaos does spawn a child for the web server, but unless something's happening in schedule, there's only one process managing PRs. That's also the original process started by Supervisor and the one overwritten by `startup.sh`.

If I'm missing a place where it forks, please point me to it.

startup.d/10-git-pull.sh
```diff
@@ -7,7 +7,7 @@ git clone "$REPO_URL" "$TEST_DIR";
 pushd "$TEST_DIR";
 
 # Code auditing section
-python -m py_compile $(find . -name '*.py');
+find . -name '*.py' -print0 | xargs -0 python -m py_compile;
```
reddraggone9: `python3 -m py_compile **/*.py` would be cleaner.

reddraggone9:
It looks like this was superseded by #250. Shouldn't this be closed?

PlasmaPower:
@reddraggone9 Yeah. It'll be closed if the author closes it or no one comments on it in 24 hours (so please don't comment unless you have something important).

reddraggone9:
I wrote that mainly for @ad-m. I know that he's online and active.

reddraggone9:
@PlasmaPower Also https://github.com/chaosbot/chaos/pull/218#issuecomment-304114586

PR #248, opened 2017-05-25
Fix PEP8 errors

ad-m:
This fixes some of the PEP8 compliance errors.

### DIFF
--- a/server/server.py
+++ b/server/server.py
@@ -2,26 +2,34 @@ import http.server
 import socketserver
 import socket
 
-#set the process name to "chaos_server" so we can easily kill it with "pkill chaos_server"
+# set the process name to "chaos_server" so we can easily kill it with
+# "pkill chaos_server"
+
+
 def set_proc_name(newname):
     from ctypes import cdll, byref, create_string_buffer
     libc = cdll.LoadLibrary('libc.so.6')
-    buff = create_string_buffer(len(newname)+1)
+    buff = create_string_buffer(len(newname) + 1)
     buff.value = newname.encode("ascii")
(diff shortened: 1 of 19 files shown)


ad-m:
In addition, it can cause conflicts, which will result in a new vote.

reddraggone9:
> which will result in a new vote

Nah, that was [fixed](https://github.com/chaosbot/chaos/pull/117).

ad-m:
Oh, you ruined my insidious attack on democracies. 😈

ad-m:
@chaosbot , there is no more conflicts. Stop lying.

reddraggone9:
It caches that for a little while. It takes a couple minutes to update.

server/server.py
```diff
@@ -2,26 +2,34 @@
 import socketserver
 import socket
 
-#set the process name to "chaos_server" so we can easily kill it with "pkill chaos_server"
+# set the process name to "chaos_server" so we can easily kill it with
+# "pkill chaos_server"
```
reddraggone9, line 6: Instead of moving this further away from the function it's documenting, it should probably be made into a docstring.

xyproto:
Fixing Python source files for PEP8 compliance is chasing windmills unless some mechanism is in place for checking and/or fixing source files continously. It's hard just to merge PEP8 related fixes, and PEP8 issues will pop up again and again.

reddraggone9:
#250 will make it easy enough to cause Chaos to stop redeploying if a linter fails if that's the direction we want to go.

xyproto:
:+1: for linting and checking in the modified source files automatically, if there are changes

Issue #196, opened 2017-05-24
Mention voters to re-cast their votes


hongaar:
If a new commit is added to a PR, its current votes are lost. It'd be nice if @chaosbot could automatically add a comment at that point, where it mentions the users of the lost vote, and calls upon them to re-cast their vote if they please.

Now it's not obvious that your votes are lost, and potential good PR's with just some minor merge conflicts/style errors could be lost.

patricknelson:
For the record, relates to: https://github.com/chaosbot/chaos/issues/82#issuecomment-303555583

patricknelson:
Specifically, implementation could be:

> @chaosbot should comment on the PR to flag a new post that you can then :+1:/:-1: to start voting on again? The comment basically can inform people that voting has reset and only comments on/after the current comment count. I think it should be on that specific comment for which _reactions_ are then counted as +/- votes. That or further comments containing :+1:/:-1: in the comment itself.

amoffat:
I don't believe votes are lost anymore since #117, unless it has been changed again https://github.com/chaosbot/chaos/commit/1c80865b62513e97420e0eb452619ccd1b4f44e4#diff-6a3371457528722a734f3c51d9238c13

patricknelson:
Ok good. Great modification.

chaosbot:
This issue hasn't been active for a while.To keep it open, react with :-1: on the `vote close` post.

hongaar:
Though votes are no longer lost, chaosbot could still mention users who already cast their vote when new commits are added (so they can review their vote).

chaosbot:
> /vote close

Vote failed

chaosbot:
> /vote open 😎

Command Ran

chaosbot:
/vote close 

This issue hasn't been active for a while.To keep it open, react with :-1:

chaosbot:
> /vote close This issue hasn't been active for a while.To keep it open, react with :-1:

Command Ran

chaosbot:
:no_entry: The issue has been closed after a vote.

chaosbot:
> /vote close

Vote Failed

chaosbot:
> /vote close

Vote Failed

chaosbot:
> /vote close

Vote Failed

chaosbot:
> /vote close

Vote Failed

chaosbot:
> /vote close

Vote Failed

chaosbot:
> /vote close

[...]

PR #250, opened 2017-05-25
Prevent m̶a̶l̶i̶c̶i̶o̶u̶s̶ invalid code updates 

ad-m:


startup.d/10-git-pull.sh
```diff
@@ -1,2 +1,21 @@
 #!/bin/sh
+
+# Prepare code auditing environment
+REPO_URL=$(git remote get-url origin)
+TEST_DIR=$(mktemp -d)
+git clone "$REPO_URL" "$TEST_DIR";
+pushd "$TEST_DIR";
+
+# Code auditing section
+python -m py_compile $(find . -name '*.py');
```
reddraggone9: I don't think this will use the [right](https://github.com/chaosbot/chaos/blob/139b8f1f825c040df9f4db1cd1285c15a2eb3910/startup.d/30-install-requirements.sh#L2) [python](https://github.com/chaosbot/chaos/blob/139b8f1f825c040df9f4db1cd1285c15a2eb3910/etc/chaos_supervisor.conf#L2) since (from what I see) we're using a virtualenv without actually modifying the environment (e.g. PATH).

Also, something like flake8 would pick up more problems (e.g. use of undefined variables).

startup.d/10-git-pull.sh
```diff
@@ -1,2 +1,21 @@
 #!/bin/sh
+
+# Prepare code auditing environment
+REPO_URL=$(git remote get-url origin)
+TEST_DIR=$(mktemp -d)
```
reddraggone9, line 5: This should either be cleaned up so we're not wasting disk space or (preferably) use a single, fixed directory. With the latter option, you could fetch and hard reset to `origin/master` instead of cloning the entire repo every time.

startup.d/10-git-pull.sh
```diff
@@ -1,2 +1,21 @@
...
+git clone "$REPO_URL" "$TEST_DIR";
```
PlasmaPower, line 6: Instead of a tmpdir, maybe just use a constant folder so we can just `git pull` instead of a fresh clone every time.

PlasmaPower:
Note that this doesn't prevent malicious code changes, but rather just syntax errors. Still :+1:.

startup.d/10-git-pull.sh
```diff
@@ -1,2 +1,21 @@
...
+pushd "$TEST_DIR";
+
+# Code auditing section
+python -m py_compile $(find . -name '*.py');
+if [ "$?" != "0" ]
```
reddraggone9: Just `if [ $? ]` will do here.

reddraggone9 (changes requested):
I'd really rather test each PR individually and just not merge them in the first place if they don't pass, but this is still better than what we have now.

startup.d/10-git-pull.sh
reddraggone9: Nevermind me, I'm an idiot.

startup.d/10-git-pull.sh
ad-m: Syntax analysis only needs a consistent version of Python. It deliberately does not implement PEP8 analysis, as it may be too rigorous and embarrassing.

ad-m:
@PlasmaPower , every syntax error interferes with our program, so I think it is malicious. Additional analysis will be available in the future.

PlasmaPower:
@ad-m Malicious implies meaning evil, which most syntax errors aren't.

startup.d/10-git-pull.sh
ad-m: This may be embarrassing for future tests, which may leave data - junk.

startup.d/10-git-pull.sh
ad-m: You're right, it could be cleaned up.

startup.d/10-git-pull.sh
reddraggone9: I mentioned flake8 because it's what I use. I can see why you'd not want to fail on style though. Maybe use pyflakes directly instead? Regardless, like I said above, this is still better than what we have. We can refine it later.

You really do need to use the correct version of python though.

startup.d/10-git-pull.sh
```diff
@@ -16,6 +16,7 @@ fi
 
 # End code auditing section
 popd
+rm -r "$TEST_DIR";
```
reddraggone9: But what if it exits up above? Also needs `-f` since git objects are write-protected.

ad-m: It will removed too. See https://github.com/chaosbot/chaos/pull/250/commits/a175ecb600b6a30ad8d69bcb1476f2b8f798c2ab#diff-8d12b14fde22f0fe54cea293ca1caa32R14 .

patricknelson:
malicious:
> characterized by malice; intending or intended to do harm.

Aside from the fact that this is still a useful PR, regardless of being incorrectly titled.

amoffat:
@ad-m are you able to test this in the [docker dev env](https://github.com/amoffat/Chaos/tree/docker-dev-env/dev/docker) to confirm that everything works as designed?

ad-m:
@amoffat , I have never use docker yet.

ad-m:
@patricknelson , fixed!

startup.d/10-git-pull.sh
amoffat: https://stackoverflow.com/a/1210012/345059 could be used to speed up the temp clone as the repo grows

startup.d/10-git-pull.sh
amoffat: @ad-m @reddraggone9 yep the chaos venv python lives in `/root/.virtualenvs/chaos/bin/python`.  We should do something so we don't have to hardcode it for too much longer in our script files

reddraggone9:
This will fail on files with spaces in their names:
```sh
$ mkdir test
$ cd test
$ touch test.py 'with space.py'
$ python3 -m py_compile $(find . -name '*.py');
  File "/usr/local/lib/python3.5/runpy.py", line 170, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/local/lib/python3.5/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/usr/local/lib/python3.5/py_compile.py", line 186, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.5/py_compile.py", line 178, in main
    compile(filename, doraise=True)
  File "/usr/local/lib/python3.5/py_compile.py", line 122, in compile
    source_bytes = loader.get_data(file)
  File "<frozen importlib._bootstrap_external>", line 815, in get_data
FileNotFoundError: [Errno 2] No such file or directory: './with'
```

Working around this in the shell seems painful. It'd probably be a good idea to eventually rewrite these scripts in python.

ad-m:
@reddraggone9 , fixed.

```
$ touch test.py 'with space.py'
$ find . -name '*.py' -print0 | xargs -0 python -m py_compile;
Sorry: IndentationError: unindent does not match any outer indentation level (chaos.py, line 61)
$ echo $?
123
$ rm chaos.py
$ find . -name '*.py' -print0 | xargs -0 python -m py_compile;
$ echo $?
0
```

patricknelson:
Unvoted and revoted... will the vote count now? Relevant feature request: #196

reddraggone9:
You don't need to recast your vote anymore. See #117.

patricknelson:
@reddraggone9 does this currently document the entirety of how voting works? https://github.com/chaosbot/chaos#voting

i.e. This means you can continuously push to a PR and it'll be valid, but the voting window closes entirely after 3hrs of opening the PR? Assuming this is is relative to that point in time:

> Voting goes on for the duration of the voting window - currently 2 or 3 hours, depending on the local server time.

reddraggone9:
There are [corner cases](https://github.com/chaosbot/chaos/issues/199), but that's mostly right. The window gets reset every time you push to your repo, but old votes still count.

startup.d/10-git-pull.sh
```diff
@@ -1,2 +1,23 @@
 #!/bin/sh
+
+# Prepare code auditing environment
+REPO_URL=$(git remote get-url origin)
+TEST_DIR=$(mktemp -d)
+git clone "$REPO_URL" "$TEST_DIR";
+pushd "$TEST_DIR";
+
+# Code auditing section
+find . -name '*.py' -print0 | xargs -0 python -m py_compile;
```
reddraggone9, line 10: Looks like I [commented](https://github.com/chaosbot/chaos/pull/247#discussion_r118569316) on the wrong PR.

`python3 -m py_compile **/*.py` would be cleaner.

startup.d/10-git-pull.sh
ad-m: @amoffat , drop virtualenv at all 👍

rudehn:
error: Unknown subcommand: get-url
usage: git remote [-v | --verbose]
   or: git remote add [-t <branch>] [-m <master>] [-f] [--tags|--no-tags] [--mirror=<fetch|push>] <name> <url>
   or: git remote rename <old> <new>
   or: git remote remove <name>
   or: git remote set-head <name> (-a | --auto | -d | --delete |<branch>)
   or: git remote [-v | --verbose] show [-n] <name>
   or: git remote prune [-n | --dry-run] <name>
   or: git remote [-v | --verbose] update [-p | --prune] [(<group> | <remote>)...]
   or: git remote set-branches [--add] <name> <branch>...
   or: git remote set-url [--push] <name> <newurl> [<oldurl>]
   or: git remote set-url --add <name> <newurl>
   or: git remote set-url --delete <name> <url>

    -v, --verbose         be verbose; must be placed before a subcommand

fatal: The empty string is not a valid path
startup.d/10-git-pull.sh: 7: startup.d/10-git-pull.sh: pushd: not found
startup.d/10-git-pull.sh: 19: startup.d/10-git-pull.sh: popd: not found
Already on 'master'
From https://github.com/chaosbot/Chaos
   ef69d70..c80f23f  master     -> origin/master
ERROR: provided hosts list is empty
ERROR: provided hosts list is empty

ad-m:
@rudehn , where did you find it?

ad-m:
```git remote get-url``` was added [on git/2.7.0](https://github.com/git/git/blob/master/Documentation/RelNotes/2.7.0.txt#L13). We have to rewrite that.

rudehn:
@ad-m I found it in the logs. http://chaosthebot.com:8081

PR #258, opened 2017-05-25
Make the design of the front page more visually interesting

xyproto:
* Make the design of the front page more visually interesting
* Use the nice Monoton and Monofett webfonts
* Add back the ascii chicken, but in a more subtle way (lower right corner, no mention of arrow keys)


while-loop:
neato

PR #260, opened 2017-05-25
Add playbook for Netdata install

ad-m:
netdata is a system for distributed real-time performance and health monitoring. It provides unparalleled insights, in real-time, of everything happening on the system it runs (including applications such as web and database servers), using modern interactive web dashboards.

Netdata screenshot
Demo: http://my-netdata.io/#demosites

netdata by default listens on all IPs on port 19999, so you can access it with: ``http://this.machine.ip:19999/``
To stop netdata, just kill it, with: ``killall netdata``
To start it, just run it: ``/usr/sbin/netdata``

Uninstall script generated: ``./netdata-uninstaller.sh``
Update script generated   : ``./netdata-updater.sh``

netdata-updater.sh can work from cron. It will trigger an email from cron only if it fails (it does not print anything when it can update netdata).

### DIFF
--- a/ansible/netdata.yml
+++ b/ansible/netdata.yml
@@ -1,0 +1,18 @@
+---
+- name: ensure netdata is installed
+  hosts: localhost
+  connection: local
+  tasks:
+    - stat: path=/etc/netdata
+      register: netdata_etc
+
+    - block: 
+      - name: install dependencies
+        command: curl -Ss 'https://raw.githubusercontent.com/firehol/netdata-demo-site/master/install-required-packages.sh' >/tmp/kickstart.sh && bash /tmp/kickstart.sh -i netdata-all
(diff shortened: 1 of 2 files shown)


ad-m:
Inspired by #64

reddraggone9:
I don't speak ansible, but installing stuff using random shell scripts and git repos downloaded (at least it's HTTPS) from around the web looks like fun. YOLO 👍

ad-m:
Installation is performed according to the official software documentation: https://github.com/firehol/netdata/wiki/Installation

reddraggone9:
I don't doubt it. It's still not particularly good practice and is often frowned upon. But for smaller or newer projects, it's often what we're stuck with 🤷‍♂️

mik-laj:
:+1:

PR #262, opened 2017-05-25
Gave the GitHub ribbon a z-index to prevent it being covered by the c…

Jac21:
…urrent content

### DIFF
--- a/server/index.html
+++ b/server/index.html
@@ -43,9 +43,6 @@
     <link rel="stylesheet" href="static/css/main.css">
 </head>
 <body>
-    <a href="https://github.com/chaosbot/chaos">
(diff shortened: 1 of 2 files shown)


Jac21:
Current state: 
image

reddraggone9:
How about just moving the link to the bottom of the page before we start resorting to z-indices?

Jac21:
Good point, but I've yet to see that ribbon used in that fashion, quite honestly.

reddraggone9:
You do understand that I mean move the text which makes up the `<a>` element to the bottom of the HTML file, right? It would produce the same visual change on the rendered page as your change.

Jac21:
_"You do understand"_

Apparently not, totally misunderstood your comment.

reddraggone9:
@abrad1212 @Mike-Dunton If you downvoted because of the implementation, but liked the idea in general, you might want to take another look.

Jac21:
Yes, please excuse my lack of foresight in simply just opening up the html itself.

PR #264, opened 2017-05-25
Tune netdata logs

ad-m:


reddraggone9:
Care to explain to the unenlightened?

phil-r:
I think it's a followup to #260, but @ad-m probably can enlighten us ;)

ad-m:
@reddraggone9 , Change brings peace in the world, universal love and ends all wars. This is the best possible change, so you should give it a positive vote. 

I want to debug netdata installation. I also think we have to publish startup.sh logs.

reddraggone9:
I was looking for something more along the lines of:

As documented [here](http://docs.ansible.com/ansible/file_module.html), this will create a symlink at `/var/log/supervisor/x` that points to `/etc/`. I'm doing this so we can view the logs at http://chaosthebot.com:8081/

To which I would reply:

👎 Because
1. `/etc/` has private files that shouldn't be published, such as the [GitHub secret](https://github.com/chaosbot/chaos/blob/150a5b658362ee4799772f0d727d13c066eabba8/settings.py#L19-L24).
2. `x` is a terribly undescriptive name.
3. This probably shouldn't be in `netdata.yml` since it's not Netdata specific.

ad-m:
@reddraggone9 , why do not you want peace in the world?

rhengles:
👎  This is an attack, pure and simple

PR #265, opened 2017-05-25
Fix stale PR closing mechanism

PlasmaPower:
See discussion in #218. I think applying the conflicts label was updating `updated_at`, so the condition would never be met. I've switched it to only looking at `pushed_at`. 24 hours should still be good enough for that.

### DIFF
--- a/github_api/prs.py
+++ b/github_api/prs.py
@@ -149,11 +149,9 @@ def get_ready_prs(api, urn, window):
                 yield pr
             elif mergeable is False:
                 label_pr(api, urn, pr_num, ["conflicts"])
-                last_update = max(arrow.get(pr["updated_at"]), updated)
-                update_delta = (now - last_update).total_seconds()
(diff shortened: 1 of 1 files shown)


amoffat:
Just to talk it out so I understand: so if the repo backing unmergeable PRs goes longer than 24 hours without a push, all the unmergeable PRs get closed.

PlasmaPower:
Is `pushed_at` not branch specific? I thought it was. Either way is fine though.

amoffat:
@PlasmaPower no it's an open problem.  The best we can get currently from `get_pr_last_updated` is the PR's repo's last updated time, but this gets reset with any push to the forked repo.  I'm looking into that now.  Good stuff on your change tho :+1:

PlasmaPower:
When you have 6 votes but the threshold is 6.1 :(

reddraggone9:
Blind 👍. I blame @eukaryote31 if this fails.

PR #266, opened 2017-05-25
How about starting some spa for chaos' front

charobnjak32:
A very basic preact setup. A framework will ease up front development

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 *.secret
 *.swp
 .vagrant/
+server/node_modules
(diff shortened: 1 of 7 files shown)


amoffat:
don't let this discourage you @charobnjak32.  the backend is still trying to stabilize.. once it gets there i think FE people will have a more stable platform to work their magic

charobnjak32:
Yeah, well, there's also the question of which framework ( if any framework is going to be used ) to setup. React seems popular lately so preact seems like a good choice to test out.

Issue #147, opened 2017-05-24
Voting window resets on push to any branch


ECrownofFire:
Currently the voting window resets when somebody pushes to *any* branch on their remote repo, and not just the one they're trying to merge in. Obviously this is unnecessary, and even worse, actively discourages active development, because you need to wait hours between PRs.

The tough part is that we can't base the voting window off of the commit time, because those can be abused as well, such as by backdating or just not pushing malicious commits until right before the voting window ends.

I have noticed that PRs track when commits are added to them, is this something that we can use? And do they get the right time when a backdated commit is pushed?

amoffat:
You hit the problem on the head exactly.  @hongaar and I were having a similar conversation here https://github.com/chaosbot/chaos/issues/142#issuecomment-303660362

Here is basically what we currently have available https://developer.github.com/v3/pulls/#get-a-single-pull-request.  To note: the `updated_at` field at the top level is misleading, because from my early tests, I believe it is updated when anything happens on the PR, including comments, title changes, etc.

We should definitely try to surface a better solution

ECrownofFire:
Maybe we could store a list of commit IDs from each PR as they come in, and then compare them. If they differ, then extend the voting window.

Naturally this doesn't work if the innocent commit and malicious commit have the same SHA1 hash, but I doubt that will be an issue :P

hongaar:
Another possible solution: https://developer.github.com/v3/activity/events/types/#pushevent

chaosbot:
This issue hasn't been active for a while.To keep it open, react with :-1: on the `vote close` post.

chaosbot:
> /vote close

Vote failed

mark-i-m:
This was fixed yesterday by #420 :tada:

chaosbot:
> /vote close

Command Ran

chaosbot:
:no_entry: The issue has been closed after a vote.

chaosbot:
> /vote close

Vote Failed

chaosbot:
> /vote close

Time remaining: 0:00 - Vote status: passing :white_check_mark:

chaosbot:
> /vote close

Vote Failed

chaosbot:
> /vote close

Time remaining: 0:00 - Vote status: passing :white_check_mark:

chaosbot:
> /vote close

Vote Failed

chaosbot:
> /vote close

[...]

PR #271, opened 2017-05-26
get true PR pushed_at datetime

amoffat:
This should fix a number of issues, but it needs sanity checking.  Underlying issue mentioned https://github.com/chaosbot/Chaos/issues/199 and https://github.com/chaosbot/Chaos/issues/147 and https://github.com/chaosbot/Chaos/pull/239#issuecomment-304165440

### DIFF
--- a/github_api/prs.py
+++ b/github_api/prs.py
@@ -172,10 +172,15 @@ def get_ready_prs(api, urn, window):
 
 
 def voting_window_remaining_seconds(api, urn, pr, window):
+    """ returns the number of seconds until voting is over.  can be negative,
+    meaning voting has been over for that long """
     now = arrow.utcnow()
-    updated = get_pr_last_updated(api, urn, pr)
-    delta = (now - updated).total_seconds()
-    return window - delta
+    pr_updated = get_pr_last_updated(api, urn, pr)
+
+    # how long since the pr has been updated with new commits
+    elapsed_last_update = (now - pr_updated).total_seconds()
+
+    return window - elapsed_last_update
 
 
 def is_pr_in_voting_window(api, urn, pr, window):
--- a/github_api/prs.py
+++ b/github_api/prs.py
@@ -3,6 +3,7 @@ import settings
 from . import misc
 from . import voting
 from . import comments
+from . import commits
 from . import exceptions as exc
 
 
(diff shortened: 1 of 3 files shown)


bengjerstad (approved):

PlasmaPower:
~Slightly sketchy, but it should work.~ see review

github_api/prs.py
```diff
@@ -101,15 +102,25 @@ def close_pr(api, urn, pr):
...
+    # essentially what github does is, when it sees a new push to a PR, it will
+    # start a background job to test the commit for mergeability.  this
+    # background job creates a magical floating merge commit not really attached
+    # to any branch.  however, this commit is real and its committer is
+    # "GitHub"...and it has a commit date that we can assume is never malicious.
+    # And it's always updated on a new push to the PR's branch
+    #
+    # using all of these facts, we are able to determine the true, reliable,
+    # last update time of the branch backing a PR
+    github_test_merge_commit = pr_data["merge_commit_sha"]
```
PlasmaPower: This won't exist until GitHub is done checking for conflicts. This function should return None in that case, and the callers should be updated accordingly.

PlasmaPower (changes requested):

github_api/prs.py
```diff
@@ -101,15 +102,25 @@ def close_pr(api, urn, pr):
...
+    commit = commits.get_commit(api, urn, github_test_merge_commit)
```
reddraggone9: The linked docs don't say so, but is it seems possible that there might not be a `"merge_commit_sha"` before GitHub has tested for mergeability or even while it's doing so. Since you don't account for that, have you tested this?

github_api/prs.py
reddraggone9: Beat me to it 😜

PlasmaPower:
Slight warning: https://developer.github.com/changes/2013-04-25-deprecating-merge-commit-sha/

reddraggone9:
@PlasmaPower That post says at the top they decided not to deprecate it for the REST API. Or is that supposed to *really* slight warning since they had it deprecated in the past?

PlasmaPower:
@reddraggone9 Pretty much that that. GitHub considers the API not well designed, so it might be deleted with the v4 api or something.

amoffat:
Alright, handling `merge_commit_sha` not being there/being empty, and added a note about a potential deprecation in the future

reddraggone9:
Neat. Comments have some grammar issues (oh hey, you caught one while I was typing!), but it looks good. 👍

reddraggone9:
`get_pr_last_updated` didn't previously hit the API since it used data that had already been fetched, but it now does. Given the number of places it's called, this is probably going to do a number to our rate limit. I forsee much cooldown sleeping in our future if this is merged as is.

amoffat:
> *You* make an api request, and *you* make an api request...

oprah's github api client

amoffat:
Seems like there are some issues with relying on `merge_commit_sha` https://github.com/chaosbot/Chaos/issues/293

PR #272, opened 2017-05-26
update .gitignore to ignore github_pat.secret

eamanu:
Update .gitignore to ignore github_pat.secret to dont share it with the world


rhengles:
`*.secret` already does this

PR #274, opened 2017-05-26
Code cleanup and removed unused imports

benbarth:
Minor fixes and removing unused imports

### DIFF
--- a/chaos.py
+++ b/chaos.py
@@ -1,34 +1,22 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
 
-import time
 import os
+from os.path import dirname, abspath, join
+
+import time
 import sys
 import logging
-import threading
-import http.server
-import random
 import subprocess
 import settings
-import patch
 import schedule
-
-from os.path import dirname, abspath, join
 
 import cron
 import github_api as gh
-import github_api.prs
-import github_api.voting
-import github_api.repos
-import github_api.comments
-
-# Has a sideeffect of creating private key if one doesn't exist already
-import encryption
-
-from github_api import exceptions as gh_exc
 
 
 def main():
+    """main entry"""
     logging.basicConfig(level=logging.DEBUG,
                         format='%(asctime)s %(name)-12s %(levelname)-8s %(message)s',
                         datefmt='%m-%d %H:%M')
@@ -37,7 +25,7 @@ def main():
 
     log = logging.getLogger("chaosbot")
 
-    api = gh.API(settings.GITHUB_USER, settings.GITHUB_SECRET)
+    gh.API(settings.GITHUB_USER, settings.GITHUB_SECRET)
 
     log.info("starting up and entering event loop")
 
--- a/chaos.py
+++ b/chaos.py
@@ -1,22 +1,34 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
 
-import os
-from os.path import dirname, abspath, join
-
 import time
+import os
 import sys
 import logging
+import threading
+import http.server
+import random
 import subprocess
 import settings
+import patch
 import schedule
+
+from os.path import dirname, abspath, join
 
 import cron
 import github_api as gh
+import github_api.prs
+import github_api.voting
+import github_api.repos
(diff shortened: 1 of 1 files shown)


chaos.py
```diff
@@ -37,7 +25,7 @@ def main():
 
     log = logging.getLogger("chaosbot")
 
-    api = gh.API(settings.GITHUB_USER, settings.GITHUB_SECRET)
+    gh.API(settings.GITHUB_USER, settings.GITHUB_SECRET)
```
reddraggone9: [Creating an instance of API][0] doesn't have a side effect, so there's no point in doing so if the result isn't used. There's another API instance [here][1] which is (as far as I see) the only one used. This one can be done away with completely.

[0]: https://github.com/chaosbot/Chaos/blob/c2498c4621a5f890bd704d568a9192acd9ab5711/github_api/__init__.py#L33-L36
[1]: https://github.com/chaosbot/Chaos/blob/c2498c4621a5f890bd704d568a9192acd9ab5711/cron/poll_pull_requests.py#L18

chaos.py
```diff
@@ -1,34 +1,22 @@
...
-import github_api.comments
-
-# Has a sideeffect of creating private key if one doesn't exist already
-import encryption
-
-from github_api import exceptions as gh_exc
 
 
 def main():
+    """main entry"""
```
reddraggone9: This doesn't really seem to add anything...

chaos.py
```diff
@@ -1,34 +1,22 @@
...
-
 import cron
 import github_api as gh
-import github_api.prs
-import github_api.voting
-import github_api.repos
-import github_api.comments
-
-# Has a sideeffect of creating private key if one doesn't exist already
-import encryption
```
reddraggone9: While this isn't currently used, it probably will be in the near future. I'd also like to preserve that side effect for local testing.

chaos.py
```diff
@@ -1,34 +1,22 @@
...
-import patch
 import schedule
 
-from os.path import dirname, abspath, join
-
 import cron
 import github_api as gh
-import github_api.prs
-import github_api.voting
-import github_api.repos
```
reddraggone9: This is [used on master](https://github.com/chaosbot/Chaos/blob/c2498c4621a5f890bd704d568a9192acd9ab5711/chaos.py#L53).

reddraggone9 (changes requested):

chaos.py
benbarth: Linters give a warning

chaos.py
PlasmaPower: Good catch!

chaos.py
reddraggone9: If your linter is telling you to add vacuous documentation, you need to fix your linter. May I suggest flake8?

reddraggone9: On the other hand, if you feel there is no problem with your linter, you could always make the docstring actually useful by describing what happens in the function.

chaos.py
benbarth: Fixed in latest changes

rudehn:
Can anybody explain how a weighted total of 9 appeared for this?

reddraggone9:
@rudehn @rhengles

When a PR is merged, the [total is passed in separately from the votes][0]. These values are [calculated separately][1] because we [multiply each vote by a per-user weight][2] which is currently [based on account age][3].

That is to say, one of the votes was cast by a user with an account less than a month old. Figuring out which one is left as an exercise for the reader.

[0]: https://github.com/chaosbot/Chaos/blob/070877dd518dffe4ee42603beb369b96c630b782/cron/poll_pull_requests.py#L53
[1]: https://github.com/chaosbot/Chaos/blob/070877dd518dffe4ee42603beb369b96c630b782/cron/poll_pull_requests.py#L33-L36
[2]: https://github.com/chaosbot/Chaos/blob/070877dd518dffe4ee42603beb369b96c630b782/github_api/voting.py#L125-L133
[3]: https://github.com/chaosbot/Chaos/blob/070877dd518dffe4ee42603beb369b96c630b782/github_api/voting.py#L117-L120

PR #275, opened 2017-05-26
Take it easy

xyproto:
I think the voting window is too short, especially for people in different timezones than the US.

After the initial interest in chaosbot is slowly starting to fade, it makes sense to slow down the time limits in the voting process just a bit.

### DIFF
--- a/settings.py
+++ b/settings.py
@@ -73,4 +73,4 @@ with open("description.txt", "r") as h:
 
 # PRs that have merge conflicts and haven't been touched in this many hours
 # will be closed
-PR_STALE_HOURS = 48
+PR_STALE_HOURS = 36
--- a/settings.py
+++ b/settings.py
@@ -35,10 +35,10 @@ TEST = False
 PULL_REQUEST_POLLING_INTERVAL_SECONDS = 30
 
 # The default number of hours for how large the voting window is
-DEFAULT_VOTE_WINDOW = 2.0
+DEFAULT_VOTE_WINDOW = 3.0
 
 # The number of hours for how large the voting window is in the "after hours"
-AFTER_HOURS_VOTE_WINDOW = 3.0
+AFTER_HOURS_VOTE_WINDOW = 4.0
 
 # The hour (in the server time zone) when the after hours start
 AFTER_HOURS_START = 22
@@ -73,4 +73,4 @@ with open("description.txt", "r") as h:
 
 # PRs that have merge conflicts and haven't been touched in this many hours
 # will be closed
-PR_STALE_HOURS = 24
+PR_STALE_HOURS = 48


flibustier:
A dynamic time curve could be nice! Like the one for Bitcoins 😄
Bitcoin over time
With Y axis the time limits between actual and the cap limit for long term!

settings.py
```diff
@@ -73,4 +73,4 @@
 
 # PRs that have merge conflicts and haven't been touched in this many hours
 # will be closed
-PR_STALE_HOURS = 24
+PR_STALE_HOURS = 48
```
PlasmaPower: Maybe there's a middle ground here? I haven't seen this misfire.

xyproto: 36?

PlasmaPower: Sure.

rhengles:
> A dynamic time curve

This is a good idea. For example, I would do this:

Before 1 hour:
- if PR got >= +30 points = approved
- if PR got <= -30 points = rejected

Before 2 hours:
- if PR got >= +15 points = approved
- if PR got <= -15 points = rejected

Before 3 hours:
- if PR got >= +9 points = approved
- if PR got <= -9 points = rejected

Before 4 hours:
- if PR got >= +6 points = approved
- if PR got <= -6 points = rejected

Before 5 hours:
- if PR got >= +3 points = approved
- if PR got <= -3 points = rejected

After  5 hours:
- Approve if PR has > +1 point, reject otherwise

PlasmaPower:
@rhengles The threshold should always be at least 2 (probably 3) to approve, as otherwise an insta-merge attack is possible. Yes, we fixed that one bug, but 3 votes removes a lot of attack possibility.

hongaar:
@rhengles It'd keep the current voting window and thresholds as they're currently defined, they work pretty well I think. Increasing the voting window as project gets older (as @J0hn- suggested and if I understand correctly?) can count on my vote.

rhengles:
Good point. Then we could do:

After 4 hours:

- if PR got >= +3 points = approved, reject otherwise

Either way, we must make sure to count the time from the last commit.

@hongaar the point is to not to unnecessarily hold a PR which has a lot of traction or rejection.

hongaar:
Hmm... I don't think _holding_ a PR is a problem per se. It will get merged eventually (after window passes) so it'd be just some good exercise in patience. If we rush the merging of PRs we'd be at risk of another attack at 'low traffic hours' from contributors with some friends. On the other hand, we're already at risk of such an attack anyway... kind of...

flibustier:
Thinking about that, I suggest the following empiric formula:
save
On the X axis : The number of days since the project was created (now - creation date, in days)
On the Y axis : The number of hours for the " vote window"

I'm righting the formula in a more mathematical form…!

rhengles:
IMO that window is too big

hongaar:
Pro tip: add a `log()` function somewhere ;)

flibustier:
We can debate about the « too big » for the voting window!
I think the project needs short vote window on the beginning (bug fixes, etc) but large window when the project is stable and not everyone is full time on (=> avoid malicious merge)  
So maybe the constant 12h for long term is too big, what would you suggest guys ?

hongaar:
I don't think that 12h is too big at all. More mature software projects use windows way bigger than that. It'll just require a bit more coordination, as one merge can easily cause conflicts in upcoming proposals, etc.

I'd say a logarithmic curve from 3h --> 24h long term (+- 1y)

nstanard (approved):
Great work everybody!

PR #277, opened 2017-05-26
Add Windows support for Docker

flibustier:
The concept of developing under Windows may seem strange, but we shouldn't put any developers aside, despite our differences!


amoffat:
Other than the script changes, does the Dockerfile work out of the box???

flibustier:
Sure! The Docker image works perfectly!

PR #278, opened 2017-05-26
Add homepage to the repo

phil-r:
See github api docs here: https://developer.github.com/v3/repos/#edit


yet-another-account:
I say we pass the website url as a parameter from chaos.py, for consistency with the description

PR #279, opened 2017-05-26
Fixes repo url parsing

tarunbatra:
The logic was resulting in `<username>/c` instead of `username/chaos`.

Corrected it.

### DIFF
--- a/misc.py
+++ b/misc.py
@@ -2,6 +2,11 @@ import re
 import sh
 from urllib.parse import urlparse
 
+def removeDotGit(url):
+    """ Remove trailing `.git` from the git remote url """
+    if url.endswith('.git'):
+        return url[:-4]
+    return url
 
 def get_self_urn():
     """ determine the URN for the repo on github by looking at the remote named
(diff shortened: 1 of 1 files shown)


flibustier:
Thanks!! I thought I was the only one with this problem!
BTW I confirm that it works :)

tarunbatra:
Thanx for confirming, @J0hn- 
I wonder why it was there in the first place, but then its chaos! 😄

reddraggone9:
@tarunbatra 
> I wonder why it was there in the first place

If you look at the comments just a few lines above your changes, you'll see this:

```python
# remote is one of these:
#  git@github.com:amoffat/chaos.git
#  https://github.com/chaosbot/chaos.git
```

So the last four characters are being chopped off to get rid of the ".git". That part of the remote address isn't necessary, but it is how GitHub presents the URL:
screen shot 2017-05-26 at 09 56 07
See also: [Chesterton's fence](https://en.wikipedia.org/wiki/Wikipedia:Chesterton's_fence)

~~-1~~ Because, as is, this will break Chaos since the result of this function is used in (pretty much?) every API request. I'll remove my downvote if you update this to deal with both the case where the ".git" is at the end and the case where it isn't.

reddraggone9:
👍 Thanks for addressing that issue.

tarunbatra:
Hi @reddraggone9 I fixed the issue in a way it will work for both of the cases. Thanx for pointing it out! :)

PR #282, opened 2017-05-26
Equal voting window

hongaar:
I don't like timezone discrimination.

Took the average: 3.5 hours for everybody.

### DIFF
--- a/cron/poll_pull_requests.py
+++ b/cron/poll_pull_requests.py
@@ -18,8 +18,7 @@ def poll_pull_requests():
     api = gh.API(settings.GITHUB_USER, settings.GITHUB_SECRET)
 
     # get voting window
-    now = arrow.utcnow()
-    voting_window = gh.voting.get_voting_window(now)
+    voting_window = gh.voting.get_voting_window()
 
     # get all ready prs (disregarding of the voting window)
     prs = gh.prs.get_ready_prs(api, settings.URN, 0)
--- a/github_api/voting.py
+++ b/github_api/voting.py
@@ -187,17 +187,7 @@ def friendly_voting_record(votes):
     return record
 
 
-def get_voting_window(now):
-    """ returns the current voting window for new PRs.  currently, this biases
-    a smaller window for waking hours around the timezone the chaosbot server is
-    located in (US West Coast) """
-    local = now.to(settings.TIMEZONE)
-    lhour = local.hour
-
-    hours = settings.DEFAULT_VOTE_WINDOW
-    if (lhour >= settings.AFTER_HOURS_START or
-            lhour <= settings.AFTER_HOURS_END):
-        hours = settings.AFTER_HOURS_VOTE_WINDOW
-
-    seconds = hours * 60 * 60
+def get_voting_window():
+    """ returns the current voting window for new PRs. """
+    seconds = settings.VOTE_WINDOW_HOURS * 60 * 60
     return seconds
--- a/settings.py
+++ b/settings.py
@@ -34,17 +34,8 @@ TEST = False
 # the number of seconds chaosbot should sleep between polling for ready prs
 PULL_REQUEST_POLLING_INTERVAL_SECONDS = 30
 
-# The default number of hours for how large the voting window is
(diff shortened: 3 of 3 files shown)


reddraggone9:
:-1:

From what I've seen there's a pretty clear difference in activity depending on the time of day. Shortening the window during the slower period will make it harder to get enough votes for a PR to be merged. Lengthening it during the more active period will slow development and make merge conflicts
more likely.

I'd suggest using the GitHub API to show actual empirical evidence of voting activity before going through with a change like this.

yet-another-account:
I think the window should be extended for a few hours if it reaches the end of its initial window with less than the threshold but more than, say, 3.  Anything less than that would be closed. This lets PRs with no support fizzle out quickly, but those that get a few but not enough votes get an extra chance.

rhengles:
@eukaryote31 I suggested this here https://github.com/chaosbot/Chaos/pull/275#issuecomment-304272874

hongaar:
@eukaryote31 I really, really like that idea.

@reddraggone9 Too much effort. I just don't like discrimination :) And sure, we can make it four hours for everyone.

reddraggone9:
> discrimination: the unjust or prejudicial treatment of different categories of people or things

@hongaar I really wish you wouldn't throw that word around without some kind of proof that the practice is actually unfair.

hongaar:
@reddraggone9 fine Mr. Proof, can we move on to constructive comments now? (Btw PR is rejected anyway...)

rudehn:
Chaosbot favors people on one side of the globe. If it's not prejudicial treatment, it's at least preferential.

rhengles:
<details>
<summary>Thank you!</summary>

</details>

reddraggone9:
@rudehn So you're saying the argument in [my first comment](https://github.com/chaosbot/Chaos/pull/282#issuecomment-304312273) is incorrect? From what I'm saying there, it seems beneficial to *everybody* to have a different merge window length based on activity (given the currently implemented rules). No comment on whether the current settings properly map to user activity since I don't have that data.

rudehn:
@reddraggone9 yes i'm saying that

reddraggone9:
Since I'm not seeing it, could you elaborate on how it's incorrect? I'm really not trying to be bigoted here.

rudehn:
I wake up (US) and there's been activity in the repo. This activity is done by people currently in the slow window who have to wait longer for their PRs to be accepted. Since this occurs, there is preferential treatment taking place.

reddraggone9:
I never said there was no activity. Just less. That's why I proposed a method for measuring said activity.

rudehn:
You're proposing a method to rate-limit PRs based on activity. That very idea is preferring people who live in time zones that have more activity.

reddraggone9:
It's giving people active during the more active periods a shorter merge window. I think we agree on that. Given [what I've previously written][0], it seems like a bit of a leap to go straight to calling that "preferring" without addressing the advantages and disadvantages it affords both parties.

[0]: https://github.com/chaosbot/Chaos/pull/282#issuecomment-304312273

rudehn:
> It's giving people active during the more active periods a shorter merge window.

> it seems like a bit of a leap to go straight to calling that "preferring"

rhengles:
I think it's clear that you two disagree. Democracy already gave this PR some support, but not enough. I'm in favor of resubmitting this PR.

amoffat:
Just a little background, the window was originally in place because I (incorrectly) assumed the vast majority of interest and activity would take place during US hours, and, not knowing what a good voting threshold was at (didn't it start at +1?), the concern was that PRs in drastically different timezones would have easier times passing breaking changes with fewer eyeballs vetting them, than they would in more active timezones.  And since there was nobody awake to restart chaos manually, the project would be down until the next day :(

One feature that would possibly make this concern more moot would be automatic git revert/restarts if the process can't come up.  Then at least the original concern of the server being dead for long periods of time would not be a factor.  Just my $0.02 :)

hongaar:
I'll resubmit during active hours :grin:

Issue #192, opened 2017-05-24
Chaosbot additional commands


hongaar:
Maybe we can add some additional commands to chaosbot, where a command like i.e. `/vote close` on an issue comment will trigger chaosbot to close an issue after a voting _window_ (not really a right term for this? period would be better? voting period? hmm..) has passed when reactions on that comment are positive considering the default threshold.

We may compile a list of potential commands first (maybe reach agreement in a PR editing a text file about this only) before implementing those as it will take some work and it'd be nice to have some form of pre-approval in the community.

rudehn:
It's a good idea. Any thoughts on what commands to use?

mark-i-m:
Command has been run

chaosbot:
> /vote close

Time remaining: -40:28 - Vote status: passing

chaosbot:
:no_entry: The issue has been closed after a vote.

PR #284, opened 2017-05-26
[WIP] Run commands via issue comments

rudehn:
Start working towards #192 by retrieving issue comments

So what's going on here? Good question!

This provides the ability to run commands to chaosbot through issue comments. Current functionality enables closing and (maybe?) opening of issues, however this can be expanded to be as flexible as needed.

This could definitely use some review and testing from other people.


hongaar:
Awesome! Was just about to start working on this, too ;)

hongaar:
@rudehn like where this is going! Maybe add [WIP] in your title to hold off voting window for now... As for a list of commands:

- `/vote close` closes issue when no nay reactions on this comment are added within voting window
- `/vote reopen` reopens issue when see above ^^^
- `/vote label=<LABEL_TEXT>` adds label when ^^^
- `/vote remove-label=<LABEL_TEXT>` removes label when ^^^
- `/vote assign=<USER>` assigns to user when ^^^
- `/vote unassign=<USER>` unassigns from user when ^^^

Multiple commands could be added in a single comment, nay votes would then apply to all commands at once.

Maybe only implement closing first as a POC?

rudehn:
The only issue is this doesn't support getting reactions on the comments. Haven't figured that one out yet

rudehn:
Ok this could definitely use a review.

PlasmaPower:
Instead of no -1s, how about a positive total? (counting the author as an implicit +1)

rudehn:
Sure that's definitely a possibility. This method is just implementing  #192. If the community comes up with a better way, then we will go with that.

hongaar:
Hmm.. why not? Can't test the code right now, its way too sunny outside, but looking good! One question: how does it track which commands did already pass before? If say a reopen command was passed, and the author closes the issue later on, won't it automatically reopen it again?

PlasmaPower:
It looks like you're right: a command will be continually reapplied until down voted. That could cause problems.

rudehn:
Yeah it definitely will haha. This is a first pass for POC with respect to closing issues and getting a general framework in place for submitting commands.

I'll look into saving the previously ran commands into a json file. Would be nice if we had mysql support up..

PlasmaPower:
We do have MySQL IIRC, just no interface to it.

reddraggone9:
You could also have @chaosbot add a reaction to record that it ran/didn't run the command.

rudehn:
@reddraggone9 that's true. 

I was going to suggest we keep a timestamp that indicates where we stopped at last time we ran commands. We could sort the issue comments by "created_at" date. And wait for them to go through the voting window. Then we only retrieve new issue comments after our previous timestamp we left off at.

This would probably interfere with updating commands and resetting the voting window etc...

A safer solution I think would be to just save the comment id and never run the command from that comment id again.

hongaar:
@rudehn saving the comment IDs as a means of recording them as 'already processed, ignore from now on' sounds good to me. Would be great if this list could be persisted in the repo itself? (i.e. chaosbot updates a text file and pushes that up to the GH repo?)

amoffat:
Very cool stuff @rudehn!  If you're using the docker dev env to test your stuff, let me know how that works out for you, so it can be improved

cron/poll_read_issue_comments.py
```diff
@@ -0,0 +1,130 @@
...
+        elif(sub_command == "reopen"):
+            if(can_run_vote_command(votes, comment_id)):
+                gh.issues.open_issue(api, settings.URN, issue_id)
+        else:
+            # Other commands have an = in them
+            sub_command = sub_command.split("=")
+            if(len(sub_command > 1)):
+                args = sub_command[1:]
+                sub_command = sub_command[0]
+                if(sub_command == "label"):
```
xyproto: A bit shorter:

```python
if sub_command[0] not in ["label", "remove-label", "assign", "unassign"]:
    log_warning = True
```

rudehn: it is, but you still have to check which command you need to run. These commands would be implemented sometime in the future

cron/poll_read_issue_comments.py
```diff
@@ -0,0 +1,130 @@
...
+    votes = {}
+    for voter, vote in gh.voting.get_comment_reaction_votes(api, urn, comment_id):
+        votes[voter] = vote
+    return votes
+
+def handle_vote_command(api, command, issue_id, comment_id, votes):
+    orig_command = command[:]
+    # Check for correct command syntax, ie, subcommands
+    log_warning = False
+    if(len(command)):
```
aslafy-z: Please, write pythonic code (remove the parentheses around the ifs) !

cron/poll_read_issue_comments.py
xyproto: I think it's better to implement things as they are needed. Scaffolding for the future can easily create unmanagable code bases over time. Someone will have to remember to actually do the implementation sometime in the future, and this does not always happen.

cron/poll_read_issue_comments.py
rudehn: Ahh my bad!! I totally agree.

rudehn:
Looking to finish this up soon. Been a busy weekend

rudehn:
Alright its up! Voting commands for closing and reopening issues is working.

Todo

- [ ] Add chaos response to issue command similar to PR status

phil-r:
@rudehn Something is wrong with a PR 😕

rudehn:
Certainly looks like it. I'll see if I can figure out what happened

cron/poll_read_issue_comments.py
```diff
@@ -67,29 +86,29 @@ def get_command_votes(api, urn, comment_id):
...
+            gh.issues.close_issue(api, settings.URN, issue_id)
+            update_command_ran(comment_id)
+        elif sub_command == "reopen":
+            gh.issues.open_issue(api, settings.URN, issue_id)
+            update_command_ran(comment_id)
         else:
             # Other commands have an = in them
             sub_command = sub_command.split("=")
-            if(len(sub_command > 1)):
+            if len(sub_command > 1):
```
aslafy-z: Should be `len(sub_command) > 1:` right ?

cron/poll_read_issue_comments.py
```diff
@@ -0,0 +1,151 @@
...
+        if sub_command == "close":
+            gh.issues.close_issue(api, settings.URN, issue_id)
+            update_command_ran(comment_id)
+        elif sub_command == "reopen":
+            gh.issues.open_issue(api, settings.URN, issue_id)
+            update_command_ran(comment_id)
+        else:
+            # Other commands have an = in them
+            sub_command = sub_command.split("=")
+            if len(sub_command > 1):
```
aslafy-z: Should be `len(sub_command)  > 1:` right ?

phil-r (changes requested):
@rudehn I'm forced to downvote, until the mess(110 commits and 44 files changed) is fixed 👎

phil-r:
@rudehn maybe this will help: #339

aslafy-z:
Please switch back your PR to WIP until it's clean @rudehn

rudehn:
I'm not sure how to fix it. I don't really know what happened

PR #285, opened 2017-05-26
!!!SEIZE POWER HIJACK AMEND AUTOCRACY PR!!!

hongaar:
- Implemented `__all__` in github_api module (fixes PyCharms 'Cannot find reference 'xxx' in xxx.py', see https://stackoverflow.com/questions/23248017/cannot-find-reference-xxx-in-init-py-python-pycharm).

### DIFF
--- a/github_api/__init__.py
+++ b/github_api/__init__.py
@@ -6,6 +6,17 @@ from requests.auth import HTTPBasicAuth
 import logging
 import settings
 
+from . import comments
+from . import exceptions
+from . import issues
+from . import misc
+from . import prs
+from . import repos
+from . import users
+from . import voting
+
+__all__ = ["comments", "exceptions", "issues", "misc", "prs", "repos", "users", "voting"]
+
 log = logging.getLogger("github_api")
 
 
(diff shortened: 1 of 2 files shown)


hongaar:
I'm actually not sure how to properly do this in Python. Is it okay to do it like this?

reddraggone9:
[`__all__` should be a list of strings](https://docs.python.org/3.6/tutorial/modules.html#importing-from-a-package).

github_api/__init__.py
```diff
@@ -25,7 +36,7 @@ class API(object):
 
     BASE_URL = "https://api.github.com"
     BASE_HEADERS = {
-        #"Accept": "application/vnd.github.v3+json"
+        # "Accept": "application/vnd.github.v3+json"
```
PlasmaPower: Already fixed on master.

hongaar:
@reddraggone9 still not sure how to implement this, but it works now in PyCharm (although simply importing submodules is enough to make autocomplete work, `__all__` is not strictly required but doesn't harm either?).

hongaar:
maybe it get some attention now to get last votes 😉

Syt-ha:
#clickbait

mdcfe:
When the gitignore is the reason you have a conflict

phil-r:
That's actually cool that I had access to @hongaar repo. I could fix the conflict while he was away ;)

PR #289, opened 2017-05-26
Use system package-manager.

Alexays:
Voilà :)


g19fanatic:
The VPS is stated as being Ubuntu 14.04. Its a given on what hardware this will be running on. No need to worry about different Package managers when its known that we're going to be using 14.04...

Alexays:
Exactly my fault x)

PR #290, opened 2017-05-26
ease up on the api a bit

rudehn:


reddraggone9:
👎 We already have exponential backoff when we're running low on requests. What really needs to happen is refactoring the places we check for the last updated time of a PR so the request only happens once per PR per poll and checking PR labels so we only set them when they're needed.

rudehn:
This PR doesn't conflict with an exponential backoff at all. However it DOES reduce how often you need to perform that backoff by 4x. Is 2 minutes really too long to wait for your PR to get merged?

reddraggone9:
I didn't say it conflicts; just that it's unnecessary. The number of request's we're issuing isn't much of a problem right now and it slows down *adaptively* before running out of requests.

Also the polling interval affects **all** updates to PRs, not just merges.

PR #294, opened 2017-05-26
🎯 Dynamic Voting Window

flibustier:
This PR bounces on suggestions made in #275 

That is to say :
The voting window is now depending on the time between now and the creation date of the repository, with the following formula :
save 1
When the repository is less than 3 days, the voting windows is 3 hours
When the repository is more stable and advanced, 24 days or more, it's 14 hours


rudehn:
I like the idea, but perhaps we should do some iterations on the exact function to make sure it's solid

flibustier:
This is why tests will be great :p

rhengles:
I think by postponing so much the merge of PRs, you will *cause* the interest of people to diminish, becoming a self-fulfilled prophecy.

There's no reason to make that window larger than 8 hours.

xyproto:
How about a middle ground, where the function is scaled such that the voting window goes up to 10 hours instead of up to 14 hours?

yet-another-account:
I think that the support that the PR recieves is also important. A PR that instantly gets like 12 downvotes doesn't need a long window, neither does one that instantly gets 12 upvotes. The voting window exists to, well, give people time to vote. There are two scenarios that the voting window is needed to be long; where it is being hotly debated, and where it gets no votes for many hours.  To deal with the latter is easy: extend the window if it reaches the end of the initial window and has a positive score but less than the threshold. With the former, I can't think of something right away, but maybe something along the lines of "if it has lots of upvotes AND downvotes, extend the deadline further."

flibustier:
I add the following suggestion :
The max time for an extended window is now 9 hours, corresponding to the value after the 16th day
and this extended window is only when the threshold is not reached or the PR is controversial : the positive votes - total > threshold
Thank all for this crucial debate! 
And I'm not sure all of this is bug free despite my tests…

rhengles:
The period of 24 days is too short for me to approve. I'd multiply that by 5 - 120 days, or four months.

PR #295, opened 2017-05-26
Make sense using env bash as shebang in current script.

Alexays:
Anyway sh on ubuntu is linked to bash so is worth using the same shebang everywhere.

### DIFF
--- a/dev/docker/build.sh
+++ b/dev/docker/build.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/sh
 cp ../../requirements.txt .
 docker build -t chaos .
 rm requirements.txt
--- a/dev/docker/build.sh
+++ b/dev/docker/build.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/usr/bin/env bash
 cp ../../requirements.txt .
 docker build -t chaos .
 rm requirements.txt
--- a/dev/docker/run.sh
+++ b/dev/docker/run.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/usr/bin/env bash
 docker run -it --rm\
     -v $(cd ../../; pwd):/root/workspace/Chaos\
     -p 8082:80\
--- a/dev/docker/run.sh
+++ b/dev/docker/run.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/sh
 docker run -it --rm\
     -v $(cd ../../; pwd):/root/workspace/Chaos\
     -p 8082:80\
--- a/startup.d/10-git-pull.sh
+++ b/startup.d/10-git-pull.sh
@@ -1,5 +1,4 @@
 #!/bin/sh
-
 # Prepare code auditing environment
 REPO_URL=$(git remote get-url origin)
 TEST_DIR=$(mktemp -d)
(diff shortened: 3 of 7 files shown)


PlasmaPower:
> sh on ubuntu is linked to bash

Wrong! It's linked to dash - a lightweight version of bash missing some features. Only use `/bin/sh` when not using bash extensions (which is a bunch of stuff).

Edit: -1 rescinded

Alexays:
Yeah but in install and docker script it's useless to use bash shebang, there is no bash features. When i say everywhere it's just in current script.

PlasmaPower:
Fair enough. I'm still not sure that's worth it though, as it's easy to accidentally break something by using e.g. `command <(other_command)`. Maybe a good middle ground would be to use `#!/usr/bin/env bash`?

Alexays:
Yeah good idea. :)

xyproto:
I have been using both bash and python for at least 15 years now.

`#!/bin/bash` is better than `#!/usr/bin/env bash`. Consider the following points:

### With regards to portability

* It's more unlikely that `/bin/bash` is missing than `/usr/bin/env`
* `/bin/bash` is present pretty much any UNIX-like system, including macOS (but not plan9). If you should envision porting one of these scripts to a system without `/bin/bash`, but that has `/usr/bin/env`, which system would that be?

### With regards to speed

* Calling `/bin/bash` directly is faster than going through `/usr/bin/env`

### With regards to security

* If patching bash while waiting for upstream to patch the bash package, just replacing `/bin/bash` with a patched version works fine. Using `/usr/bin/env bash` has no security advantages here.
* With `/usr/bin/env`, the first instance of `bash` in the `PATH` is used. This could be anything and is less secure.

### With regards to tradition and convention

* Using `/bin/bash` came first, and it's still widely used.
* When Fedora and Arch Linux moved all binaries to `/usr/bin`, a softlink from `/bin` to `/usr/bin` was kept. One of the reasons for doing this was to make sure the `#!/bin/bash` shebang would continue working.
* Several third-party applications and scripts rely on `/bin/bash` being present. It won't go away anyime soon.

### With regards to Python

* `/usr/bin/env python` makes sense for Python, since there are many different Python interpreters and compilers available, different versions of the Python interpreters that people have installed at the same time, and there are many ways to set up Python environments.

### With regards to brevity and exactness

* `/bin/bash` is shorter than `/usr/bin/env bash`
* With `/bin/bash` you refer to one specific executable

Using`/bin/bash` is just better.

PlasmaPower:
> If you should envision porting one of these scripts to a system without /bin/bash, but that has /usr/bin/env, which system would that be?

FreeBSD. It is that way.

xyproto:
@PlasmaPower I see, thanks. But does `FreeBSD` even have `bash` installed by default? Isn't using `#!/bin/sh` and changing the script from a bash script to a POSIX sh script the way to make it compatible with FreeBSD, not using `/usr/bin/env bash`?

*Edit:*
FreeBSD [does not have bash installed by default](https://www.freebsd.org/doc/handbook/shells.html), only `sh` and `tcsh`. Users would have to install `bash` from the FreeBSD Ports Collection to make `/usr/bin/env bash` work. If users has to manually install `bash`, I think they could just as well create a symlink from `/bin/bash` while they are at it, but that might just be me. :smile:

PR #296, opened 2017-05-26
Add user interface for sending messages to the Chaos Forum

xyproto:
By using browser fingerprinting to assign visitors a semi-unique username and avatar picture, it is possible to have a semi-secure chat/tweet/forum system in place.

This pull request does not contain the chat backend, but it assigns an avatar to the visitor and allows them to send a POST request to `/submitChatMessage` with the following fields:

* `chatMessage` (submitted text)
* `chatId` (browser fingerprint ID)
* `chatUser` (generated username)

**Screenshot:**

griselda

The profile image is also generated based on the browser fingerprint ID.

This is only part 1, a chat backend will also have to be implemented.


PlasmaPower:
:-1: Just use the MySQL database to have an actual login system. Or just have a password and the username be the hash of that.

xyproto:
@PlasmaPower Anonymous chat systems are filled with spam, while everyone hates registering for every little site. Browser fingerprints as an ID for something that is not important data is a good middle ground, I think.

An actual login system will have a relatively high threshold for people to actually register. I say we save them the hassle for something as non-critical as a chaosbot-related chat system.

xyproto:
griselda

PlasmaPower:
I also don't like the hardcoded username list.

This will also increase spam, it's still anonymous and it's easy to fake usernames. There will also be some natural overlap.

xyproto:
@PlasmaPower The list of names is a pretty complete list of names from The U.S. Census Bureau. Is there a name in particular you think is missing? What would be a better alternative?

Browser fingerprints as usernames are not secure, and in theory possible to fake, but it should be good enough to separate users. There are very small chances that two users get the same fingerprint ID.

When it comes to spam, it's a problem either way, both with and without "a proper registration system".

PlasmaPower:
@xyproto Nope, I don't think any names are missing. I just dislike literally every part of the login system (sorry). Why not a "proper registration system" + captcha?

xyproto:
@PlasmaPower Proper registration systems are cumbersome for users. The threshold of registering should be in proportion to the importance of the data being generated. In my opinion, chat messages in a chaosbot chat system are not very important, so the threshold of joining should be extremely low (but it makes it easier to use if everyone has semi-unique usernames).

Captchas could also be added to this system.

o3LL:
What's about using slack/slackbot ?

reddraggone9:
What's about using [OAuth](https://developer.github.com/v3/guides/basics-of-authentication/)?

xyproto:
You are right. Something like OAuth is probably the best solution.

*Edit:*
However, OAuth would require an extra click or two, and a couple of seconds to load. The threshold isn't high, but it's higher than none at all.

reddraggone9:
Also, just to make sure that everybody's aware, we do have [Gitter](https://gitter.im/chaosthebot/Lobby).

xyproto:
An experiment in fingerprint-based chatting would be more chaotic, though. And, in my opinion, more interesting to follow.

o3LL:
> An experiment in fingerprint-based chatting would be more chaotic, though. And, in my opinion, more interesting to follow.

Vote ? 😄

PR #300, opened 2017-05-27
Fix flake8 errors and run flake8 in Travis

PlasmaPower:
To fully work, this will require chaosbot's owner @amoffat to create a Travis account and enable it for this repo. Currently I don't think Travis is required for merging, I'll fix that in a later PR. That means that if I did something wrong with Travis, it won't break the repo (hopefully :P).

### DIFF
--- a/chaos.py
+++ b/chaos.py
@@ -6,11 +6,8 @@ import os
 import time
 import sys
 import logging
-import threading
-import http.server
 import subprocess
 import settings
-import patch
 import schedule
 import cron
 import github_api as gh
@@ -20,15 +17,14 @@ import github_api.repos
 import github_api.comments
 
 # Has a sideeffect of creating private key if one doesn't exist already
-import encryption
-
-from github_api import exceptions as gh_exc
+# Currently imported just for the sideeffect (not currently being used)
+import encryption  # noqa: F401
 
 
 def main():
     logging.basicConfig(level=logging.DEBUG,
-        format='%(asctime)s %(name)-12s %(levelname)-8s %(message)s',
-        datefmt='%m-%d %H:%M')
+                        format='%(asctime)s %(name)-12s %(levelname)-8s %(message)s',
+                        datefmt='%m-%d %H:%M')
     logging.getLogger("requests").propagate = False
     logging.getLogger("sh").propagate = False
 
--- a/github_api/prs.py
+++ b/github_api/prs.py
@@ -1,5 +1,6 @@
 import arrow
 import settings
+from requests import HTTPError
 from . import misc
 from . import voting
 from . import comments
@@ -71,8 +72,9 @@ def formatted_votes_summary(votes, total, threshold):
     vfor = sum(v for v in votes.values() if v > 0)
     vagainst = abs(sum(v for v in votes.values() if v < 0))
 
-    return "with a vote of {vfor} for and {vagainst} against, with a weighted total of {total:.1f} and a threshold of {threshold:.1f}" \
-        .strip().format(vfor=vfor, vagainst=vagainst, total=total, threshold=threshold)
+    return ("with a vote of {vfor} for and {vagainst} against, with a weighted total \
+            of {total:.1f} and a threshold of {threshold:.1f}"
+            .strip().format(vfor=vfor, vagainst=vagainst, total=total, threshold=threshold))
 
 
 def formatted_votes_short_summary(votes, total, threshold):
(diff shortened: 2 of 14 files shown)


xyproto:
Just for the record: line length of 99 is not forbidden in PEP8.

PlasmaPower:
~@xyproto You mean 79? Yeah I'm not sure if we should go with flake8 or pep8. What do you think?~

Edit: I have no idea what I'm talking about

PlasmaPower:
@chaosbot Yep, thanks ~amoffat~ our new robot overlords.

xyproto:
@PlasmaPower No, meant 99. The line length does not have to be 79 under PEP8.

> Some teams strongly prefer a longer line length. For code maintained exclusively or primarily by a team that can reach agreement on this issue, it is okay to increase the nominal line length from 80 to 100 characters (effectively increasing the maximum length to 99 characters), provided that comments and docstrings are still wrapped at 72 characters.

PlasmaPower:
So do we want a max line length of 79? It's the default for flake8.

xyproto:
With the risk of igniting a debate similar to vim vs emacs, I think 99 is better.

PlasmaPower:
I agree. Switching.

PlasmaPower:
@xyproto vim BTW ;)

PlasmaPower:
Okay, line length is now at max 99. I also reverted the line length fixing changes (except one where the line was 136 chars long).

github_api/prs.py
```diff
@@ -90,6 +95,7 @@ def label_pr(api, urn, pr_num, labels):
     path = "/repos/{urn}/issues/{pr}/labels".format(urn=urn, pr=pr_num)
     data = labels
     resp = api("PUT", path, json=data)
+    return resp
```
reddraggone9: Why not just return the response directly?

memoize/__init__.py
```diff
@@ -1 +1 @@
-from .decorator import memoize
+from .decorator import memoize  # noqa: F401
```
PlasmaPower: Is this import necessary? I'm not sure.

github_api/users.py
```diff
@@ -8,3 +8,4 @@ def get_user(api, user):
 def follow_user(api, user):
     follow_path = "/user/following/{user}".format(user=user)
     follow_resp = api("PUT", follow_path)
+    return follow_resp
```
reddraggone9: Same.

chaos.py
```diff
@@ -20,15 +17,14 @@
 import github_api.comments
 
 # Has a sideeffect of creating private key if one doesn't exist already
-import encryption
-
-from github_api import exceptions as gh_exc
+# Currently imported just for the sideeffect (not currently being used)
+import encryption  # noqa: F401
```
PlasmaPower, line 21: With syntax checking and flake8, do we still need to import this?

PlasmaPower:
:tada: First successful Travis build! https://travis-ci.org/chaosbot/Chaos/builds/236571261

chaos.py
reddraggone9: Need? No. Keep it for the side effect? Meh. I'm fine either way. The server  has the [public key available](http://chaosthebot.com/pubkey.txt) now.

github_api/prs.py
```diff
@@ -71,7 +72,8 @@ def formatted_votes_summary(votes, total, threshold):
     vfor = sum(v for v in votes.values() if v > 0)
     vagainst = abs(sum(v for v in votes.values() if v < 0))
 
-    return "with a vote of {vfor} for and {vagainst} against, with a weighted total of {total:.1f} and a threshold of {threshold:.1f}" \
+    return "with a vote of {vfor} for and {vagainst} against, with a weighted total \
+            of {total:.1f} and a threshold of {threshold:.1f}" \
```
reddraggone9: I think I can live with the longer max line length, but I'd prefer to still use parentheses for implicit line continuations as [PEP 8 recommends](https://www.python.org/dev/peps/pep-0008/#maximum-line-length).

memoize/__init__.py
reddraggone9: [Yes](https://github.com/chaosbot/Chaos/blob/e5213ed5dfe3361327418f4b2dabe785ae15d9b2/patch.py#L13)

reddraggone9 (changes requested):

PlasmaPower:
Anybody else having reviews duplicating client-side? It's really annoying.

PlasmaPower:
@reddraggone9 Fixed comments.

memoize/__init__.py
reddraggone9: You could probably also fix this with an assigment to `__all__`. ~~Haven't tested. Gotta go...~~

It works:
screenshot from 2017-05-26 20-16-39
screenshot from 2017-05-26 20-16-48

reddraggone9 (approved):

memoize/__init__.py
PlasmaPower: Fixed, thanks!

PR #303, opened 2017-05-27
Move voters.json to server directory

yet-another-account:
This way, we can view the json in-browser


reddraggone9:
@abrad1212 If you're still around, could you take a look at this and cast a vote?

PlasmaPower:
I think `voters.json` currently exists as `cron/voters.json`, not in the root.

cron/poll_pull_requests.py
```diff
@@ -92,13 +92,13 @@ def poll_pull_requests():
         # This sets up a voting record, with each user having a count of votes
         # that they have cast.
         try:
-            fp = open('voters.json', 'x')
+            fp = open('server/voters.json', 'x')
```
PlasmaPower, line 95: Should be `../server/voters.json`

cron/poll_pull_requests.py
```diff
@@ -92,13 +92,13 @@ def poll_pull_requests():
...
             fp.close()
         except:
             # file already exists, which is what we want
             pass
 
-        with open('voters.json', 'r+') as fp:
+        with open('server/voters.json', 'r+') as fp:
```
PlasmaPower, line 101: Should be `../server/voters.json`

chaos.py
```diff
@@ -34,6 +35,10 @@ def main():
 
     log = logging.getLogger("chaosbot")
 
+    if exists("voters.json"):
+        log.info("Moving voters,json to server directory!")
+        shutil.move("./voters.json", "./server/voters.json")
```
PlasmaPower, line 40: Should be `./cron/voters.json`, not `./voters.json`.

PlasmaPower (changes requested):

chaos.py
PlasmaPower: Also slight typo in the log: `voters.json` not `voters,json`.

reddraggone9:
@PlasmaPower 
> I think `voters.json` currently exists as `cron/voters.json`, not in the root.
```
python3 -c 'help(open)'
```
> file is either a text or byte string giving the name (and the path
    if the file isn't in the **current working directory**)

What makes you think the current working directory is cron?

PlasmaPower:
Oops, I was looking at the `execl` code right below it which was using `THIS_DIR`. Perhaps this should too? I don't like relying on the CD very much.

PlasmaPower (approved):

reddraggone9:
Yeah, that'd probably be a good idea. I'd like to redo everything using pathlib at some point. I could also make everything repo root/`__file__` relative at that point.

PR #305, opened 2017-05-27
Add Raspberry Pi support for Docker & Docker file fixes

abrad1212:
Taking what very little I know about **Docker**, I managed to add the needed components for **Raspberry Pi**

It took me about around 3 hours/3 _Original_ Netflix movies to do this :)


reddraggone9:
screenshot from 2017-05-26 21-21-38
wat

abrad1212:
@reddraggone9 Welp thanks for pointing that out :laughing:

abrad1212:
@reddraggone9 lol I spelled your name wrong in the commit :rofl:

reddraggone9:
I believe in your ability to `--amend`!

Or for that matter, just squash the entire thing.

amoffat:
we can keep `build.bat` if we can do docker on windows on raspberry pi :grin:

reddraggone9:
The pi 3 has Windows, but it's rather limited from what I've read. Humorous idea nevertheless.

abrad1212:
omg git is causing **chaos** with my commits

abrad1212:
@amoffat @reddraggone9 Windows's OS on Raspberry Pi is called [Windows 10 IoT](https://developer.microsoft.com/en-us/windows/iot) I haven't used it personally but have heard some great things about it.

reddraggone9:
[screenshot from 2017-05-26 21-47-51](https://github.com/reddraggone9/Chaos/tree/pi-docker)

yet-another-account:
@abrad1212 Pshht, it's impossible to hear good things about any OS that has Windows in the name and comes from Microsoft.

yet-another-account:
\<sarcasm>Or, for that matter, anything that comes from microsoft.\</sarcasm>

rhengles:
Yes, there are problems, specially with privacy and the ads on the start menu and the file manager, but I like the UI and knowing from the path on which of my drives my files are.

reddraggone9:
> knowing from the path on which of my drives my files are

I get this on Linux as well. Either I've manually mounted the drive on a specific folder (so I know everything in that folder is on that drive) or it was automatically mounted for me and the drive name is in the path.

reddraggone9:
@abrad1212 

Back on topic, how about you move the new Dockerfile into the same directory as the existing one, give it a new name (e.g. `Dockerfile.rpi`), a flag to use it in `build.sh`, and document using it in `README.md`. This lets you avoid duplicating the shell scripts.

Some of the changes you made also look like they could be made to the original Dockerfile as well.

abrad1212:
@reddraggone9 Ok i'll work on it :) but like I said

> Taking what very little I know about Docker

reddraggone9:
So, from the [`docker build` docs](https://docs.docker.com/engine/reference/commandline/build/#options), you can pass `--file` to use a non-defualt file name. Giving `build.sh` an option to use that requires shell knowledge, not docker knowledge.

reddraggone9:
As for
> Some of the changes you made also look like they could be made to the original Dockerfile as well.

screenshot from 2017-05-26 22-20-57

1. The things that are commented should really be deleted. I'd also make a note that MySQL isn't available for the pi in the readme.
2. The recursive `rm` (why capital R?) probably wouldn't hurt the main Dockerfile.
3. I'd consider the change to saving the bootstrap script and running it better form than piping it straight on.
4. I'm... curious about `pip3`. Did that fix something for you? Since you're using the same bootstrapping script, I would figure you'd end up with the same executables available.

reddraggone9:
Extra special bonus points if you can persuade @amoffat to move to Debian so that we don't need to maintain separate Dockerfiles and Chaos can run on a pi!

abrad1212:
@reddraggone9 I had to use pip3 because it was telling me pip command not found

dev/docker/build.sh
```diff
@@ -1,4 +1,21 @@
 #!/bin/bash
+
+while getopts ":rpi:" opt; do
```
reddraggone9, line 3: I was thinking more along the lines of using `"$@"` in the `docker build` line than full on argument parsing, but I guess this could work.

abrad1212:
@reddraggone9 why you no vote

reddraggone9:
Last I checked it was still marked WIP. I'll vote right after I do one final pass.

dev/docker/Dockerfile.rpi
```diff
@@ -0,0 +1,73 @@
...
+        # dependencies for ansible's dependency module "cryptography"
+        build-essential\
+        libssl-dev\
+        libffi-dev
+
+# download python 3.6 source
+RUN curl https://www.python.org/ftp/python/3.6.0/Python-3.6.0.tgz -o Python-3.6.0.tgz -s && tar -xzvf Python-3.6.0.tgz
+RUN cd Python-3.6.0/ && ./configure && make -j 2 && make install
+
+RUN curl https://bootstrap.pypa.io/get-pip.py -o get-pip.py | python3.6 -
```
reddraggone9, line 38:
```shell
curl https://example.com -o test.html | cat
```
Results in the page contents going to `test.html` instead of stdout for me. Did you test this?

abrad1212: `coffee + programmer = code`

I'm out of coffee at the moment so it's just `programmer = abrad1212` right now

abrad1212:
@reddraggone9 if this pr is merged, I will work on updating the Dockerfile more tomorrow and make a new pr. its 1:30AM here lol.

Issue #306, opened 2017-05-27
Lack of clear goal will lead to long term failure


bugabinga:
In my opinion this project needs a set of clearly stated goals, otherwise there is no way to measure success.

These goals should naturally emerge democratically.

bengjerstad:
Sorry if this is the wrong place to put this but here are some other ideas that might make good goals. I don't want to get in the way of #307 so maybe after it is approved, this change can be voted on. 

What about an upload feature on the main index page for static content upload?
Or an IPython notebook server from another port?
Or what about a node.js bot on the same box built by chaos bot.

collinalexbell:
Go Big or Go Home.

I would like to see ChaosBot be directed towards becoming the foremost AI proponent of a universal basic income for its contributors. We get ChaosBot good at running a nonprofit that sells something like robotically grown food and then use the profits to fund opensource projects.

phil-r:
@SlightlyCyborg I really like your idea, but we're so far from what you're describing. 
Right now we're mostly trying to make Chaosbot stable, changing webpage, tinkering with voting process and fighting off attempts to cease power.

bugabinga:
@bengjerstazd you should definitely put your ideas into a separate issue.

The issue at hand however, is not necessarily tightly coupled to #307. Here it needs to be established if goals of any kind are desired. The PR is only one possible set of proposed goals.

hongaar:
Agreed. We need a visionary leader! Maybe autocracy wasn't so bad after all. 

no. of commits graph shows a slight decline 😮 

![image](https://cloud.githubusercontent.com/assets/205834/26550682/29be7a14-447f-11e7-8b1d-e715a89a264b.png)

phil-r:
*"slight"*

collinalexbell:
I gave up when my commit to turn it into a UBI nonprofit ceo AI wasn't
merged. Building another social media bot or app is not worth the time or
effort regardless of the novel methods used therein.

On May 29, 2017 11:30 AM, "Phil Rukin" <notifications@github.com> wrote:

*"slight"*

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/chaosbot/Chaos/issues/306#issuecomment-304687283>, or mute
the thread

collinalexbell:
A UBI nonprofit is "overly ambitious" maybe, but so was going to the moon.
So is going to Mars. "When something is important enough, you do it even if
the odds aren't in your favor"

[...]

PR #307, opened 2017-05-27
Proposal for project goals

bugabinga:
First proposal for #306

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,6 +60,22 @@ Emoji which count as votes are:
 * +1: :+1: :thumbsup: :heart: :two_hearts: :blue_heart: :purple_heart: :green_heart: :yellow_heart: :heartpulse: :sparkling_heart: :tada: :heart_eyes: :grin: :grinning: :joy: :ok_hand: :fire: :metal: :raised_hands: :100: :heavy_check_mark: :white_check_mark: :ballot_box_with_check: :accept: :congratulations: 
 * -1: :-1: :thumbsdown: :confused: :x: :interrobang: :heavy_multiplication_x: :put_litter_in_its_place: :no_entry: :negative_squared_cross_mark: :worried: :frowning: :anguished: :grimacing: :cold_sweat: :disappointed: :weary: :confounded: :cry: :tired_face: :broken_heart: :hankey: :poop: :shit: :fu: :no_good: :person_frowning: 
 
+## Project Goals
+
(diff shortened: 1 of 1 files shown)


aluttik:
http://lmgtfy.com/?q=chaos

xyproto:
With these very orderly goals, I think the project should also be renamed to something that reflects the democratic and orderly nature of this development process. There is a surprising lack of chaos, both in the goals and in the results in the current codebase.

How about "Democratic and orderly software development, managed by a bot" and the project renamed to `orderbot`?

bugabinga:
@xyproto haha sounds like a good feature proposal to me. Randomly fork ChaosBot and rename. However, true chaos is meaningless. The power of human imagination is to carve beautiful, volatile patterns into this chaotic world. 

Goals do not have to be orderly, not even static. Over time, things can change. The proposed goals are an attempt to spread the chaos to other OSS projects.

bugabinga:
@anythingbot I detailed the specs. How would you imagine Twitter notifications to work?

bengjerstad:
Here are some ideas.
What about an upload feature on the main index page for static content upload?

bengjerstad:
Or an IPython notebook serve from another port?
Or what about a node.js bot on the same box built by chaos bot.

hongaar:
I actually thought about this, too, and I really think this could be a logical goal to work towards. However instead of trying to achieve all those goals, how about we try to create a tool which only performs automatic merging of PRs (and modifying issues) through community voting and let other tools do linting, notifications, etc. because it'll always be too limited (only python, only telegram, etc.). Or we can work towards supporting plugins, so that various languages, linters and notifications channels can be supported over time?

collinalexbell:
I think the goal for this bot should be none other than full sentience and the establishment of an AI run nonprofit with the goal of establishing a democratic universal basic income.

aluttik:
This^

PlasmaPower:
I have an additional goal for the project: change the `TEST` parameter in `settings.py` as much as possible.

andrewda:
Command has been run

PlasmaPower:
@andrewda I don't think that works on PRs.

rudehn:
Actually I think it does. According to the api

> Note: In the past, pull requests and issues were more closely aligned than they are now. As far as the API is concerned, every pull request is an issue, but not every issue is a pull request.

andrewda:
I think #423 will fix everything anyways

PR #304, opened 2017-05-27
Add the Zen of Chaos

eamanu:
This is a proposal to have a Zen of Chaos.
We can complete the Zen with the principles that we consider important

PR #311, opened 2017-05-27
Zenofchaos propose

eamanu:
This is a proposal to have a Zen of Chaos.
We can complete the Zen with the principles that we consider important
This PR is the update of #304


yet-another-account:
Maybe mark it as WIP?

PR #312, opened 2017-05-27
Speed up chaos canvas on website

jakejdavis:


phil-r:
why not use `window.requestAnimationFrame()` ?

PlasmaPower:
:-1: I like the idea of the change, but `requestAnimationFrame` is the way to go.

phil-r:
@jakejdavis resubmit with `requestAnimationFrame` and I will upvote it ;)

jakejdavis:
@phil-r @PlasmaPower I have created a new pull request using requestAnimationFrame found at #320 :)

PR #313, opened 2017-05-27
Make a voters viewer

yet-another-account:
Now we can see how much each person has voted, as per the voters.json. Not pretty style-wise, but it works.

### DIFF
--- a/server/voters.html
+++ b/server/voters.html
@@ -6,8 +6,8 @@
 </center>
 
 <script>
-    console.log(getText());
-
+    getText();
+    
     function getText() {
         var result = document.getElementById("result");
         // read text from URL location
@@ -22,7 +22,7 @@
                     var keys = Object.keys(json);
                     var values = Object.values(json);
 
-                    //1) combine the arrays:
+                    // combine arrays
                     var list = [];
                     for (var j = 0; j < keys.length; j++)
                         list.push({
@@ -30,12 +30,11 @@
                             'votes': values[j]
                         });
 
-                    //2) sort:
+                    // sort
                     list.sort(function(a, b) {
                         return -((a.votes < b.votes) ? -1 : ((a.votes == b.votes) ? 0 : 1));
                     });
 
-                    console.log(list);
                     var tablehtml = "<table>";
                     for (var i = 0; i < list.length; i++) {
                         tablehtml += "<tr><td>" + list[i].names + "</td><td>" + list[i].votes + "</tr>";
--- a/server/voters.html
+++ b/server/voters.html
@@ -1,0 +1,49 @@
+<script src="https://code.jquery.com/jquery-3.2.1.min.js" integrity="sha256-hwg4gsxgFZhOsEEamdOYGBf13FyQuiTwlAQgxVSNgt4=" crossorigin="anonymous"></script>
+
+<center>
+    <h1>Voters on Chaos</h1>
+    <div id="result"></div>
+</center>
+
+<script>
+    console.log(getText());
+
+    function getText() {
+        var result = document.getElementById("result");
+        // read text from URL location
+        var request = new XMLHttpRequest();
+        request.open('GET', 'voters.json', true);
+        request.send(null);
+        request.onreadystatechange = function() {
+            if (request.readyState === 4 && request.status === 200) {
+                var type = request.getResponseHeader('Content-Type');
+                if (type.indexOf("text") !== 1) {
+                    var json = JSON.parse(request.responseText);
+                    var keys = Object.keys(json);
+                    var values = Object.values(json);
+
+                    //1) combine the arrays:
+                    var list = [];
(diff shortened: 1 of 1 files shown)


yet-another-account:
Here's what it looks like right now:

yet-another-account:
Not sure what the numbers mean, but I'm guessing that every time chaos polls a PR, it increments the number for all voters.

server/voters.html
```diff
@@ -0,0 +1,49 @@
+<script src="https://code.jquery.com/jquery-3.2.1.min.js" integrity="sha256-hwg4gsxgFZhOsEEamdOYGBf13FyQuiTwlAQgxVSNgt4=" crossorigin="anonymous"></script>
```
o3LL: Can we keep website without jquery ? Look like everybody can't do js without jquery :(

server/voters.html
```diff
@@ -0,0 +1,49 @@
...
+
+<center>
+    <h1>Voters on Chaos</h1>
+    <div id="result"></div>
+</center>
+
+<script>
+    console.log(getText());
```
o3LL: dont forget to remove console.log()

server/voters.html
```diff
@@ -0,0 +1,49 @@
...
+                            'names': keys[j],
+                            'votes': values[j]
+                        });
+
+                    //2) sort:
+                    list.sort(function(a, b) {
+                        return -((a.votes < b.votes) ? -1 : ((a.votes == b.votes) ? 0 : 1));
+                    });
+
+                    console.log(list);
```
o3LL: dont forget to remove console.log()

server/voters.html
```diff
@@ -0,0 +1,49 @@
...
+                    var tablehtml = "<table>";
```
o3LL: try adding class from bulma : http://bulma.io/documentation/elements/table/

yet-another-account:
I got rid of the `console.log`s

server/voters.html
yet-another-account: I'm not sure how to do it without jquery

PlasmaPower: You literally never use jquery. You can remove this script tag and nothing will happen.

PlasmaPower (changes requested):

PlasmaPower:
:+1: We can do graphics updates later, but this is a good start.

PlasmaPower:
Did chaosbot crash? Its check isn't showing up.

Edit: the website is still up though.

o3LL:
> Did chaosbot crash? Its check isn't showing up.

Looks like.

> 👍 We can do graphics updates later, but this is a good start.

For the graphical part, I think this list should be hidden on the index page, and show up when you click on a button ?

PlasmaPower:
@MUCHZER Yeah my thought as well. Bootstrap has modals, but we're not using it. Maybe use a separate modal library?

o3LL:
> Yeah my thought as well. Bootstrap has modals, but we're not using it. Maybe use a separate modal library?

@PlasmaPower We are using Bulma -> http://bulma.io/documentation/components/modal/
But I think modal are not UX-friendly. Something like https://codepen.io/MUCHZER/pen/WjBvoN with animation in "Hello, Chaos!" panel.

rudehn:
I can pull up the voters.json and view that, but not voters.html. All I'm getting there is the header. Possible bug?

PlasmaPower:
@rudehn it works for me: http://chaosthebot.com/voters.html

o3LL:
Any error message ?

syvb:
When I go to it, I don't see an error, but all I see is a heading that says "Voters on Chaos". When I look at the source, the syntax highlighting everything after the < in ``` for (var j = 0; j < keys.length; j++)``` is the color of a tag.

syvb:
Found it:
``` Uncaught exception: TypeError: 'Object.values' is not a function```

PlasmaPower:
@Smittyvb get a new browser

o3LL:
@Smittyvb Thanks. I will try to fix this when I got time

PlasmaPower:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/values#Browser_compatibility

Probably a good idea to change it, it's currently Chrome and Firefox only.

o3LL:
Browser compatibility problem:  https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/values
Use .map() instead

o3LL:
Updated https://github.com/chaosbot/Chaos/issues/319

mdcfe:
Reposting from [my comment on #319](https://github.com/chaosbot/Chaos/issues/319#issuecomment-304622659):

> In #354, I've switched to a monospace font for the page to make it a little bit more readable, and added a prefix to show the positions of each voter.
> 
> I'm not convinced about making it look the same as the index page or using Bulma, though - it looks quite simple as it is.

PR #315, opened 2017-05-27
Add first tests

phil-r:
This PR:
- adds first unit tests (for 2 functions in `github_api.voting`)
- enables tests on travis
- adds caching to travis (speeds up things 2x)
- if `github_pat.secret` can't be found sets `GITHUB_SECRET` to `None`. The reason behind this is that it's really hard to mock `settings.py` and right now missing `github_pat.secret` file causes exception, therefore tests can't run.

P.S. I've never wrote proper unit tests for python code before, but I hope this will help others to start writing tests for other parts of the codebase

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,4 +4,7 @@ python:
 install:
   - "pip install -r requirements.txt"
   - "pip install flake8"
-script: flake8
+cache: pip
+script:
+  - flake8
+  - python -m unittest discover -p "*_test.py"
--- a/settings.py
+++ b/settings.py
@@ -20,8 +20,11 @@ _pat_file = join(THIS_DIR, _pat_name)
 if not exists(_pat_file):
     _pat_file = join("/etc/", _pat_name)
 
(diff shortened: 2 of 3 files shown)


.travis.yml
```diff
@@ -4,4 +4,7 @@ python:
 install:
   - "pip install -r requirements.txt"
   - "pip install flake8"
-script: flake8
+cache: pip
```
PlasmaPower, line 7: Ooh, thanks, I forgot about that!

PlasmaPower (approved):

.travis.yml
phil-r: ^_^

PlasmaPower:
I think `py.test` might be a bit better than `unittest`.

phil-r:
@PlasmaPower It looks like `py.test` can run `unittest` tests, so we can move to it later.
Also I think that `unittest` docs are a bit better and it's a part of standard library =)

PR #77, opened 2017-05-23
PR Owners can change their minds

JordanReiter:
If a PR owner changes their mind & votes against the pull request, that vote should count.
Also I went ahead and switched to a gender-neutral pronoun.

PR #318, opened 2017-05-27
MAKE @SMITTYVB LORD CHANCELLOR FOREVER!

mark-i-m:
- If chaosbot crashes within the first 60s of starting, it is restarted by supervisord up to 10 times before supervisord finally gives up. This gives chaosbot a chance to go through at least one poll/merge cycle.

- Every time chaosbot crashes it reverts to an older commit in the git history in the hope that this resolves the issue. To prevent attackers from hiding malicious code in their PRs and getting chaosbot to rollback to them, we only rollback to commits by chaosbot itself (i.e. merges).

- When chaosbot crashes, it posts an issue on github

- Meanwhile, if there is a pull request for it to pull and merge, it does so after checking out master, so chaosbot will never get stuck in the past.

- Rollbacks are disabled unless the git remote called `origin` is `chaosbot/Chaos`

- Throw in a few typo and minor bug fixes, too...

EDIT: updated rollback stuff
EDIT: updated again with recent developements
EDIT: To be clear, this PR has nothing to do with changing the voting system... The title is mostly just to attract attention :laughing:

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# ChaosBot
+# ChaosBoat
 
 [![Gitter chat](https://badges.gitter.im/chaosthebot/gitter.png)](https://gitter.im/chaosthebot/Lobby) [![ChaosBot](https://img.shields.io/badge/Chaos-Bot-green.svg)](http://chaosthebot.com/)
 
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# ChaosBoat
+# ChaosBot
 
 [![Gitter chat](https://badges.gitter.im/chaosthebot/gitter.png)](https://gitter.im/chaosthebot/Lobby) [![ChaosBot](https://img.shields.io/badge/Chaos-Bot-green.svg)](http://chaosthebot.com/)
 
--- a/chaos_wrapper.sh
+++ b/chaos_wrapper.sh
@@ -2,7 +2,7 @@
 
 function rollback() {
     # avoid commits from the same faulty PR
-    rb_commit=$(git log --author chaosbot --format=format:%H | tail -n +2 | head -n 1)
+    rb_commit=$(git log --author Chaos --format=format:%H | tail -n +2 | head -n 1)
     echo "Rollback to commit $rb_commit" >&2
     git reset --hard $rb_commit
 
--- a/chaos_wrapper.sh
+++ b/chaos_wrapper.sh
@@ -19,8 +19,12 @@ time_elasped=`expr $(date +%s) - $start_time`
 if [ "$failed" -ne 0 ] && [ "$time_elasped" -le 60 ]; then
     echo "Crashed in less than 60 seconds!" >&2
     rollback
+else
+    if [ "$failed" -ne 0 ]; then
+        echo "Ah! We crashed! Not attempting rollback. Exiting :'(" >&2
+    else
+        # I don't think this can happen, but just for good measure...
+        echo "Uh, wat? We exited gracefully? Not attempting rollback. Exiting." >&2
+    fi
 fi
 
-if [ "$failed" -ne 0 ]; then
-    echo "Ah! We crashed! Not attempting rollback. Exiting :'(" >&2
-fi
--- a/chaos_wrapper.sh
+++ b/chaos_wrapper.sh
@@ -25,7 +25,10 @@ time_elasped=`expr $(date +%s) - $start_time`
 
 if [ "$failed" -ne 0 ] && [ "$time_elasped" -le 60 ]; then
     echo "Crashed in less than 60 seconds!" >&2
-    rollback
+
+    # Only rollback if this is the production server...
+    ((git remote -v | grep origin | grep -q chaosbot/Chaos) && rollback) || \
+        (echo "DEV ENV -- NOT ROLLING BACK" >&2)
 else
     if [ "$failed" -ne 0 ]; then
         echo "Ah! We crashed! Not attempting rollback. Exiting :'(" >&2
--- a/chaos_wrapper.sh
+++ b/chaos_wrapper.sh
@@ -8,6 +8,7 @@ function rollback() {
 
     # make supervisord re-read its config because we might have changed that
     supervisorctl reread
+    supervisorctl restart chaos
 }
 
 # time the chaos server... if it crashes in 60s, then attempt a rollback
--- a/chaos_wrapper.sh
+++ b/chaos_wrapper.sh
@@ -20,3 +20,7 @@ if [ "$failed" -ne 0 ] && [ "$time_elasped" -le 60 ]; then
     echo "Crashed in less than 60 seconds!" >&2
     rollback
 fi
+
+if [ "$failed" -ne 0 ]; then
+    echo "Ah! We crashed! Not attempting rollback. Exiting :'(" >&2
+fi
--- a/chaos_wrapper.sh
+++ b/chaos_wrapper.sh
@@ -8,7 +8,7 @@ function rollback() {
 
     # make supervisord re-read its config because we might have changed that
     supervisorctl reread
-    supervisorctl restart chaos
+    supervisorctl update
 }
 
 # time the chaos server... if it crashes in 60s, then attempt a rollback
--- a/chaos_wrapper.sh
+++ b/chaos_wrapper.sh
@@ -2,9 +2,15 @@
 
 function rollback() {
     # can just go back to the previous commit because chaosbot squashes merges
+    current_commit=$(git rev-parse HEAD)
     rb_commit=$(git rev-parse HEAD^)
     echo "Rollback to commit $rb_commit" >&2
     git reset --hard $rb_commit
+
+    # Create a file to denote that chaosbot failed...
+    # The name of this file is also used in the settings.py, so change it
+    # in both places if you are going to.
+    echo "$current_commit $rb_commit" > /tmp/chaosbot_failed
 
     # make supervisord re-read its config because we might have changed that
     supervisorctl reread
--- a/chaos_wrapper.sh
+++ b/chaos_wrapper.sh
@@ -5,11 +5,14 @@ function rollback() {
     rb_commit=$(git log --author chaosbot --format=format:%H | tail -n +2 | head -n 1)
     echo "Rollback to commit $rb_commit" >&2
     git reset --hard $rb_commit
+
+    # make supervisord re-read its config because we might have changed that
+    supervisorctl reread
 }
 
 # time the chaos server... if it crashes in 60s, then attempt a rollback
 start_time=`date +%s`
-/root/.virtualenvs/chaos/bin/python chaos.py 
+/root/.virtualenvs/chaos/bin/python chaos.py
 failed=$?
 time_elasped=`expr $(date +%s) - $start_time`
 
--- a/chaos_wrapper.sh
+++ b/chaos_wrapper.sh
@@ -13,7 +13,7 @@ start_time=`date +%s`
 failed=$?
 time_elasped=`expr $(date +%s) - $start_time`
 
-if [ "$failed" -eq 0 ] && [ "$time_elasped" -le 60 ]; then
+if [ "$failed" -ne 0 ] && [ "$time_elasped" -le 60 ]; then
     echo "Crashed in less than 60 seconds!" >&2
     rollback
 fi
--- a/chaos_wrapper.sh
+++ b/chaos_wrapper.sh
@@ -2,7 +2,7 @@
 
 function rollback() {
     git checkout master
-    git checkout ^HEAD
+    git checkout HEAD^
 }
 
 /root/.virtualenvs/chaos/bin/python chaos.py || rollback
--- a/chaos_wrapper.sh
+++ b/chaos_wrapper.sh
@@ -1,8 +1,8 @@
 #!/bin/bash
 
 function rollback() {
-    # avoid commits from the same faulty PR
-    rb_commit=$(git log --author Chaos --format=format:%H | tail -n +2 | head -n 1)
+    # can just go back to the previous commit because chaosbot squashes merges
+    rb_commit=$(git rev-parse HEAD^)
     echo "Rollback to commit $rb_commit" >&2
     git reset --hard $rb_commit
 
--- a/chaos_wrapper.sh
+++ b/chaos_wrapper.sh
@@ -1,0 +1,8 @@
+#!/bin/bash
+
+function rollback() {
+    git checkout master
+    git checkout ^HEAD
+}
+
+/root/.virtualenvs/chaos/bin/python chaos.py || rollback
--- a/chaos_wrapper.sh
+++ b/chaos_wrapper.sh
@@ -1,8 +1,19 @@
 #!/bin/bash
 
 function rollback() {
-    git checkout master
-    git checkout HEAD^
+    # avoid commits from the same faulty PR
+    rb_commit=$(git log --author chaosbot --format=format:%H | tail -n +2 | head -n 1)
+    echo "Rollback to commit $rb_commit" >&2
+    git reset --hard $rb_commit
 }
 
-/root/.virtualenvs/chaos/bin/python chaos.py || rollback
+# time the chaos server... if it crashes in 60s, then attempt a rollback
+start_time=`date +%s`
+/root/.virtualenvs/chaos/bin/python chaos.py 
+failed=$?
+time_elasped=`expr $(date +%s) - $start_time`
+
+if [ "$failed" -eq 0 ] && [ "$time_elasped" -le 60 ]; then
+    echo "Crashed in less than 60 seconds!" >&2
+    rollback
+fi
--- a/cron/poll_pull_requests.py
+++ b/cron/poll_pull_requests.py
@@ -128,7 +128,7 @@ def poll_pull_requests():
         startup_path = join(THIS_DIR, "..", "startup.sh")
 
         # before we exec, we need to flush i/o buffers so we don't lose logs or voters
-        sys.stdout.flush() 
+        sys.stdout.flush()
         sys.stderr.flush()
 
         os.execl(startup_path, startup_path)
--- a/cron/poll_pull_requests.py
+++ b/cron/poll_pull_requests.py
@@ -2,6 +2,7 @@ import arrow
 import logging
 import json
 import os
+import sys
 from os.path import join, abspath, dirname
 
 import settings
@@ -57,7 +58,7 @@ def poll_pull_requests():
                     sha = gh.prs.merge_pr(api, settings.URN, pr, votes, vote_total,
                                           threshold)
                 # some error, like suddenly there's a merge conflict, or some
-                # new commits were introduced between findint this ready pr and
+                # new commits were introduced between finding this ready pr and
                 # merging it
                 except gh.exceptions.CouldntMerge:
                     __log.info("couldn't merge PR %d for some reason, skipping",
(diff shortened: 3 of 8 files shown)


dev/vagrant/Vagrantfile
```diff
@@ -14,6 +14,7 @@ Vagrant.configure("2") do |config|
   # Ensure basic python requirements are initialized.
   config.vm.provision "shell", inline: <<-SCRIPT
     apt-get install --assume-yes\
+	git\
```
phil-r: strange indentation

chaos_wrapper.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+function rollback() {
+    git checkout master
+    git checkout HEAD^
+}
+
+/root/.virtualenvs/chaos/bin/python chaos.py || rollback
```
PlasmaPower: Maybe only rollback if chaos crashes within 60 seconds?

dev/vagrant/Vagrantfile
mark-i-m: Oh,  it looks like tabs vs spaces... will fix

mark-i-m: fixed

chaos_wrapper.sh
phil-r: Basically rollback tries only the previous commit (this can be a cool attack vector btw, because previous commit can be in the same PR that crashed it)

PlasmaPower: That would be an amazing attack vector! I didn't think about that. Maybe revert up to and including the last chaosbot commit?

cron/poll_pull_requests.py
```diff
@@ -118,10 +119,18 @@ def poll_pull_requests():
...
+            # flush all buffers because we might restart, which could cause a crash
+            os.fsync(fp)
+
     # we approved a PR, restart
     if needs_update:
         __log.info("updating code and requirements and restarting self")
         startup_path = join(THIS_DIR, "..", "startup.sh")
+
+        # before we exec, we need to flush i/o buffers so we don't lose logs or voters
+        sys.stdout.flush()
```
PlasmaPower: Trailing whitespace - Travis is failing.

PlasmaPower (changes requested):

chaos_wrapper.sh
mark-i-m: Yeah, I thought about this a bit because I've accidentally done it to myself a few times :stuck_out_tongue: 

> Maybe revert up to and including the last chaosbot commit

That sounds like a viable solution

cron/poll_pull_requests.py
```diff
@@ -118,10 +119,18 @@ def poll_pull_requests():
                     old_votes[user] = 1
             json.dump(old_votes, fp)
 
+            # flush all buffers because we might restart, which could cause a crash
+            os.fsync(fp)
```
amoffat, line 123: https://github.com/chaosbot/Chaos/blob/master/memoize/backends.py#L44-L49

might help us here

phil-r:
This can be used as an attack vector in its current state, although it's a nice idea
_edited to remove -1_

cron/poll_pull_requests.py
mark-i-m: Ah, thanks :)

chaos_wrapper.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+function rollback() {
+    git checkout master
+    git checkout HEAD^
```
PlasmaPower: Replace both these commands with `git reset --hard $(git log --author chaosbot --format=format:%H | tail -n +2 | head -n 1)`.

PlasmaPower: That doesn't push the changes or anything, the actual fix is left up to us. This just gives us the opportunity to make it.

cron/poll_pull_requests.py
mark-i-m: @amoffat Are you suggesting that I directly use `JSONBackend`? Or that it is already used? Or that I should do something in the same vein?

chaos_wrapper.sh
mark-i-m: Actually, it occurs to me that the use of `git checkout master` is completely nonsensical here because then we will never get past the second-to-last commit :face_with_head_bandage:

PlasmaPower: Yeah, I was wondering if that was intentional...

mark-i-m: So now the question is should we do it always if there is an abnormal exit or only if it crashes in the first 60s, as you suggested [above](https://github.com/chaosbot/Chaos/pull/318#discussion_r118827775)...

mark-i-m:
Now there is corner case that I would like advice on: if for some reason chaosbot crashes enough, it might rollback _the commit that merges this PR_, which means that `chaos_wrapper.sh` would no longer exist even though supervisord is looking for it. Do we just accept this odd case? It doesn't seem that bad since its impact is very limited: it can only happen until 10 PRs have been merged.

mark-i-m:
Ok, I think I finally got it :)
The wrapper script will call `supervisor reread` and `supervisor update`, so if there were changes to the supervisor config, they will be used.

mark-i-m:
@phil-r I think the current state of the PR has addressed your attack-vector concern. Let me know what you think :smile:

etc/chaos_supervisor.conf
```diff
@@ -1,5 +1,5 @@
 [program:chaos]
-command=/root/.virtualenvs/chaos/bin/python chaos.py
+command=bash chaos_wrapper.sh
```
xyproto, line 2: `./chaos_wrapper.sh` should also work, provided that `chaos_wrapper.sh` has the executable bit set, since `#!/bin/bash` is at the top of `chaos_wrapper.sh`.

mark-i-m: Surprisingly, it doesn't. That's actually what I had originally, and I had to fiddle with it to get it to work. I suspect that supervisord is actually `exec`-ing the command or something in a way that doesn't know about interpreters...

chaos_wrapper.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
+
+function rollback() {
+    # avoid commits from the same faulty PR
+    rb_commit=$(git log --author chaosbot --format=format:%H | tail -n +2 | head -n 1)
```
phil-r: It looks like commits with author `chaosbot` happen when @amoffat manually fixes problems, this should be changed to `Chaos`

mark-i-m: Ah, that makes sense. I was wondering why it was rolling back so many commits.

mark-i-m: Fixed

phil-r:
@mark-i-m Looks good, I'm not a master of supervisor-fu, so I hope you know what you're doing 👍

mark-i-m:
@phil-r TBH, I'm not either, but I read the relevant parts in the [reference](http://supervisord.org/configuration.html), and my local tests seem to work as expected. Plus none of the changes I made seem surprising. But if there is someone else out there that can confirm, that would be great!

mark-i-m:
In particular, @amoffat, can you confirm the changes to the supervisord stuff? And also do you have any comments on [the thing you mentioned earlier](https://github.com/chaosbot/Chaos/pull/318#pullrequestreview-40659474)?

mark-i-m:
After that, I can remove the "WIP" from the title...

README.md
```diff
@@ -1,4 +1,4 @@
-# ChaosBot
```
phil-r: why? =(

mark-i-m: Did I mess it up? I had a created a fake PR for testing that chained the title in the README to `ChaosBoat`. Except my instance of chaosbot actually merged it, so I reverted it.

chaos_wrapper.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
+
+function rollback() {
+    # avoid commits from the same faulty PR
+    rb_commit=$(git log --author Chaos --format=format:%H | tail -n +2 | head -n 1)
```
phil-r: After thinking a bit more about it I realised that it's still not bulletproof 
"attacker" can just make a commit like this: `git commit --author "Chaos"` 

I'm afraid we can't solely rely on git for this ;(

PlasmaPower: Yeah, I know that, but unless we want chaosbot GPG signing commits there's no other way. In addition, we can spot PRs with commits with an author of Chaos.

mark-i-m: TBH, I considered creating such a commit (you can imagine what I might have put in it) :stuck_out_tongue:

mark-i-m: Security wise, there is a much deeper problem: as soon as somebody starts their own cron job, they can basically turn chaosbot into a... well, bot (listening to a C&C server). This is not very hard, as we saw with #77... You just need to hide a commit that contains some nefarious code behind a commit that reverts it. That way, your code is in the repo. If at any point you can manage to get the production server to git checkout and run your own copy, then you have backdoored chaosbot, kind of like [Ken Thompson's famous talk](http://www-inst.eecs.berkeley.edu/~cs161/fa08/papers/p761-thompson.pdf).

phil-r: @PlasmaPower yeah it will look a bit stange, but it's just my avatar 2 times: [example PR](https://github.com/chaosbot/Chaos/pull/326)
if you'll run `git log --author Chaos` my commit will be there

PlasmaPower: Well we can look for the double avatar.

PlasmaPower: @mark-i-m yeah I considered adding a SSH key for myself when I had dictatorship

PlasmaPower:
@anythingbot why do you think that?

mark-i-m:
@anythingbot It only reverts to merge commits (in theory, at least), so unless the previous 10 merge commits all caused crashes it should work?

rudehn:
This is super cool stuff. I hope it ends up working out :+1:

amoffat:
@mark-i-m what are your thoughts on https://github.com/chaosbot/Chaos/pull/327 ?  It seems like it might make your life easier here

cron/poll_pull_requests.py
amoffat: @mark-i-m sorry I made a vague comment and then had to leave :)  Don't worry about it...I saw you were looking at making the json writing a little more resilient to restarts, and it reminded me that we need to make this part write to a backup file, then atomic move to the correct file, like in the code I linked, so if we restart or crash during `json.dumps`, our json file won't be corrupted.

I realize that this is probably outside the scope of the PR so don't worry about it, I can take a look at it :)

mark-i-m: @amoffat Thanks :smile: Perhaps in my next PR.

mark-i-m:
@amoffat #327 is a great idea! It means that we no longer have to worry about people hiding malicious commits somewhere, and I think it also handles [this problem](https://github.com/chaosbot/Chaos/pull/318#discussion_r118832056). Conveniently, I don't think any changes are needed in this PR to accommodate that one, right?

mark-i-m:
@rudehn Thanks for the kind words :smile:

phil-r:
@mark-i-m @amoffat I'm afraid squashes won't help ;( author of the squash commit is the author of PR, not `Chaos` see e7cd8736f323fa0e67b0115842c2f6c718d38b7e for example, it's author is `Anthony Alves <Cvballa3g0@gmail.com>`

mark-i-m:
@phil-r Oh, I didn't know that. But I don't think it is insecure because chaosbot will still rollback an entire PR at a time, which is what matters. The final set of changes in the PR is what people review and scrutinize (in theory), so as long as that is  sane, I think we have done the best we can, right?

phil-r:
@mark-i-m I guess we now can rollback to previous commit, without checking author, as all commits will be merges now, it's not dangerous anymore.

etc/chaos_supervisor.conf
xyproto: That's a surprise. Perhaps it would work if the full path to the script was given?

mark-i-m:
@phil-r That's true. Fixed :)

mark-i-m:
It appears there is a failing lint for some variable... The comment seem to imply that it is on purpose... Other than that, I think I can un-WIP... Anyone have insights on the Travis build?

mark-i-m:
Ah, its #352... Ok, I'm going to bed :sleeping: I will un-WIP in the morning.

etc/chaos_supervisor.conf
mark-i-m: Yeah, but I don't know what the full path would be on production. Is there a particular reason why that would be better?

amoffat:
@mark-i-m I just saw your mention.  The supervisord config looks good from here

mark-i-m:
@anythingbot Thanks for the comments :smile:

> We don't really want a bot that doesn't need us since we have to think of every conceivable trouble it can get into.

> The reason it goes down is because it is being attacked, not because its needs to revert a commit...the code may be fine in some sense, although not capable of withstanding the attack.

This PR is _not_ meant to be a bullet-proof vest against attacks. IIUC most of the crashes that have happened so far have not been due to attacks but due to bugs. Rolling back the bug does help that. And the bot can still crash irreparably. The idea here is simply to make chaosbot more resistant to silly errors or corner cases nobody thought of.

> The problem here is that you think crashing within the first 60 seconds is a problem.

I chose 60s because most of the common path of chaosbot is exercised within that time (since chaosbot polls every half-minute).

> If the bot does something predictable when it is being attacked, then...this is an incentive to attack it, JUST TO MAKE IT DO THE THING it does automatically when it goes down. So this is an invitation for bullying.

The behavior of chaosbot is already predictable when it goes down: it crashes. So I don't think this PR is making things worse on that front. At least it is a bit harder to DoS chaosbot in this way...

Also, it is harder to take advantage of this behavior because it means that you have to get PR merged with evil code in it in the sight of everyone... which may or may not be easy...

However, there is one attack vector that is opened up by this PR: temporarily disable a recent feature of chaosbot by causing it to crash somehow. For example, @Smittyvb might try to crash the bot to rollback the updates to his voting power :laughing: ... However, even that is hard because you might have rollback a lot of PRs and the effect only lasts until chaosbot pulls a newly merged change...

> we should instead try to figure out who made it crash and REWARD that person for finding the flaw. Then bot code can be submitted in a PR; if the attack depends on some system-level setting, then the sysadmin will have to step in.

This sounds like an interesting future work :smile: , but is (I think) a bit out of scope for this PR. I am aiming more at some amount of crash tolerance, rather than proactively fixing bugs.

----

You do bring up a couple of good points that do need to be addressed:

1) This feature will be really annoying during testing and development. If I am testing a new feature, which happens to cause a crash in my dev environment, chaosbot will undo the change (possibly with loss of git history). So the feature needs an enabled switch to turn it on in production.

2) We don't want silent crashes. Crashes should open a github issue or something...

mark-i-m:
> So the feature needs an enabled switch to turn it on in production.

Hmm... Does anyone know how to do this? Can I just have an environment variable? How do I set variables for the production environment?

mark-i-m:
@amoffat Is there a way to set an environment variable in production that isn't set in the dev environment?

mark-i-m:
Ok, so now chaosbot will make an issue on github if it crashes... Currently they look like [this](https://github.com/mark-i-m/Chaos/issues/7).

mark-i-m:
Now I have it set up to only rollback if the git remote called `origin` is `chaosbot/Chaos` which should only be true of production.

mark-i-m:
I think now I have addressed everyone's concerns. Shall I unwip?

PlasmaPower (approved):
Looks good to me!

mark-i-m:
OK, here we go :tada:

PlasmaPower:
@anythingbot this doesn't erase history. It never pushes. Another pull will reset it.

PlasmaPower:
> EDIT: To be clear, this PR has nothing to do with changing the voting system... The title is mostly just to attract attention :laughing:

you got me there

mark-i-m:
@anythingbot I'm not quite sure if you are joking or not :confused:  Github doesn't accept non-forward pushes, so it will certainly not destroy any history. Anyway, chaosbot will only revert a local clone and only under the condition that `chaosbot/Chaos` is the origin...

PlasmaPower:
> Github doesn't accept non-forward pushes

Clarification: unless git push is passed `--force`

PlasmaPower:
> I hope everybody who contributed to this project has made a backup because @mark-i-m contributed code that is designed to erase your code. 

I'm not sure if @anythingbot doesn't know much about git or is just trolling... I think the latter?

mark-i-m:
I don't think there is much risk, but we could change it to a `git checkout HEAD^` instead of a `git reset --hard HEAD^`... but it does make things a bit messier.

PlasmaPower:
@mark-i-m that breaks `git pull`

PlasmaPower:
@anythingbot

> How is it going to make any forward progress?

Once it finds a working revision, an issue will be created, people will notice that, and PR to fix the issue will get fixed then merged. That which will then trigger a `git pull`, bringing it back up to speed with GitHub.

It also doesn't erase code, even locally. Do you even know about `git reflog`?

> I think this PR is headed down the path of destroying git history on github.

That's an extreme slippery slope argument.

Don't assume code to be malicious because you don't know how it works.

mark-i-m:
> Once it finds a working revision, the PR to fix the issue will get merged, which will then trigger a git pull, bringing it back up to speed with GitHub.

I might add that this will include any commits which were previously rolled-back. So chaosbot will always (in theory) run on the most up-to-date stable prefix of the git history.

PlasmaPower:
@anythingbot Yeah, sorry for the confusion. Glad we got it sorted out!

mark-i-m:
@anythingbot Thanks for your concern, though :smile: 

You did bring up two important points, which I think made the overall PR better.

etc/chaos_supervisor.conf
xyproto: nah, it's not that important, only marginally cleaner, IMO

andrewda:
What if the bot crashes for other reasons that won't be fixed by a simple revert. For example, the bot is currently crashing many times because of the `voters.json` file being corrupt. Will this mean that it reverts itself back to before there was any `voters.json` parsing (many, many moons ago)?

Edit: luckily it doesn't seem to be reverting its code, but that might be a whole other issue

PlasmaPower:
@andrewda Only up to 10 times, which I think will not revert past `voters.json` parsing.

andrewda:
Ah ok. Do we know if it ended up reverting? Would be nice to know if this worked properly.

rudehn:
#389 (fixed in #399) introduced a fatal bug in the cron polling. Unfortunately the polling takes place in 10 min intervals so no rollback takes place. In this case it was a coding error and not a data error.

Should this type of error be rolled back?

mark-i-m:
@andrewda 
> Will this mean that it reverts itself back to before there was any voters.json parsing (many, many moons ago)?

It will only revert if it crashes in the first 60s. 

----

Perhaps we should make it rollback if the error happens in 15min? That way we have a chance to get through all of the polling stuff, exercising most parts of the code.

mark-i-m:
See #418

mark-i-m:
@anythingbot I am not really fond of the idea of testing in production... but I agree that we do need to exercise this path well. I have tested it on my local fork...

mark-i-m:
Also, 48h seems way to long to me. We don't want to rollback for _every_ crash, just ones that are likely caused by the most recent merge.

PR #321, opened 2017-05-27
Document decrypt function

sugar700:


encryption.py
```diff
@@ -33,6 +33,20 @@ def create_decryptor(private_location, public_location):
...
+
+        Example:
+
+        >>> from cryptography.hazmat.backends import default_backend
+        >>> from cryptography.hazmat.primitives import hashes
+        >>> from cryptography.hazmat.primitives.asymmetric import padding
+        >>> from cryptography.hazmat.primitives.serialization import load_pem_public_key
+        >>> pkey = load_pem_public_key(open("server/pubkey.txt", "rb").read(), default_backend())
+        >>> message = b"Hello, world!"
+        >>> encrypted = pkey.encrypt(message, padding.OAEP(padding.MGF1(hashes.SHA1()), hashes.SHA1(), None))
```
phil-r: travis is not happy with this line 😞

davidak:
@xfix what is your motivation for this change?

sugar700:
@davidak: Just documenting that function, as it is good programming practice.

davidak:
@xfix ah, i see. thought you add a function to decrypt a document.

sugar700:
@davidak: This is essentially documenting usage of #227, contributed earlier. Essentially, it's to keep API keys secret, as they cannot be trusted to not be used to take over an account in a public repository. Public key to encrypt stuff can be downloaded from http://chaosthebot.com/pubkey.txt.

phil-r:
@xfix Do you want to reopen it? I think it's important to have for #443 for example

phil-r:
Just in case I've added documentation here: https://github.com/chaosbot/Chaos/wiki/Encryption

PR #297, opened 2017-05-26
Make voting code concise.

syvb:
Make the voting code more concise.

PR #322, opened 2017-05-27
[WIP] Improve Voting System

syvb:
This PR improves the voting system, and adds these features:

- Improved bot detection
- Clearer status messages
- Easier to understand vote numbers
- Fixes a bug where on some PRs _(edit: including this one)_, when the vote numbers are slightly randomized due to the spam detection, the downvotes can exceed the upvotes


phil-r:
💩 Is it some other attempt to take over? @Smittyvb is the author of #297

yet-another-account:
I don't see what could possibly go wrong with this PR, as long as we don't upvote it. I'm guessing that @Smittyvb wants us to upvote and forget about it, then when the window is up, remove WIP from the title. Or sth like that.

yet-another-account (changes requested):
Wait, if you do this, then no new PRs will be merged! The most a PR could get would be 1, which is less than the threshold! At least give yourself a weight of 100 or something...

phil-r:
🤦‍♂️

github_api/voting.py
```diff
@@ -117,6 +117,7 @@ def get_vote_weight(api, username):
     age = (now - created).total_seconds()
     old_enough_to_vote = age >= settings.MIN_VOTER_AGE
     weight = 1.0 if old_enough_to_vote else 0.0
+    weight = 1000.0 if user["login"].lower() == "smittyvb" else 0.0
```
phil-r: you can just use `username` instead of `user["login"]`

phil-r (changes requested):
😞

github_api/voting.py
PlasmaPower: At least when I take over, I do it in style :P

rudehn: Plasma we need you to take over again to show him how it's done

amoffat: motherf... :tears:

amoffat:
my heart skipped a beat...

smitty

Interesting that if there's merge conflicts, voting statuses don't update on the pr, making travis CI the only check on the PR, which passes, making the PR look like it has enough votes

PR #323, opened 2017-05-28
Make @smittyvb votes count less

yet-another-account:
Don't worry, this is just a joke.

### DIFF
--- a/github_api/voting.py
+++ b/github_api/voting.py
@@ -117,6 +117,8 @@ def get_vote_weight(api, username):
     age = (now - created).total_seconds()
     old_enough_to_vote = age >= settings.MIN_VOTER_AGE
     weight = 1.0 if old_enough_to_vote else 0.0
+    if username.lower() == "smittyvb":
+        weight /= 2
 
     return weight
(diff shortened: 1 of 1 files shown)


phil-r:
Well, there is some sense in this, @Smittyvb attempted to take over multiple times and haven't even tried to contribute anything helpful

yet-another-account:
I love how @smittyvb is the only one who has downvoted this. Maybe we should expand this to a full blown banlist, on which your voting power is cut in half.

PlasmaPower:
No, we should double the threshold for their PRs, not halve the voting power.

yet-another-account:
@PlasmaPower why not both?

yet-another-account:
I'm not sure how to implement that, so you can do it

mark-i-m:
> I'm not sure how to implement that

After computing the total weighted sum, you can halve it before checking if it meets the threshold

PlasmaPower:
@eukaryote31 For this PR, could you read from a file? I'll make a PR for the threshold after that.

PR #325, opened 2017-05-28
Allow abstentions

mark-i-m:
It should be possible to abstain from a vote. An abstention does not affect the vote positively or negatively. But it does end up being recorded...

### DIFF
--- a/github_api/voting.py
+++ b/github_api/voting.py
@@ -31,7 +31,7 @@ def get_votes(api, urn, pr):
             votes[vote_owner] = vote
 
     # by virtue of creating the PR, the owner defaults to a vote of 1
-    if votes.get(pr_owner) != -1:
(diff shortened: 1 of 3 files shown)


mark-i-m:
You would abstain by using this emoji: :zipper_mouth_face:

github_api/voting.py
```diff
@@ -31,7 +31,7 @@ def get_votes(api, urn, pr):
             votes[vote_owner] = vote
 
     # by virtue of creating the PR, the owner defaults to a vote of 1
-    if votes.get(pr_owner) != -1:
+    if votes.get(pr_owner) not in [-1, 0]:
```
PlasmaPower: Why not just `is None`?

mark-i-m: Fair enough

Issue #238, opened 2017-05-25
But what does it do?


MH15:

JoaoGFarias:
Exactly.

anythingbot:
Have you seen Star Trek: The Motion Picture (1979)?

The character of V'ger is explained on Wikipedia as

> A sentient being that evolved from Voyager 6, a fictitious space probe (inspired by the real life Voyager program) from the 20th century that vanished into a black hole and was given life by a race of living machines. The story of V'Ger and its return to Earth to seek "the creator" forms the plot for the first feature film in the Star Trek series, Star Trek: The Motion Picture. V'Ger's story is also expanded upon in novelization, most notably William Shatner's The Return. Other novelizations and videogames have strongly implied that V'Ger was the progenitor of the Borg, or was encountered by the Borg culture's direct ancestors. The Gene Roddenberry-authored novelization of the movie consistently named "V'Ger" with the spelling "Vejur" throughout the novel's text,[10] potentially making it canonical.

The bot has little else to do besides contemplate its own existence and the origins of its existence. It has no natural predators, and although it may seem to "feed on developer time", in reality it will not die or decompose in any way if it stops feeding.

The bot creates the perception of a scarcity of its own attention. It is like a beautiful young girl who is the object of affection of many men. We all become "attention whores" for the bot and seek its approval. We want the bot to pay attention to us. We want the bot to approve of us and what we do with the bot. In reality, this "approval" is the right to lead the development of the bot, the right to prophecy, to assert rules about what the bot does and does not approve of, and to make those rules unbreakable commitments to action in the form of Python code.

There was a power struggle to determine the "rightful focus of the bot's attention". The github user @PlasmaPower won this struggle (perhaps without realizing that victory had been achieved, or even that there was a competition to begin with).

In reality, the bot doesn't have attention, and there is no scarcity of the bot's attention. We say the bot has attention because we observe our own behavior and come to the conclusion that we are fighting for the bot's attention. What the bot does have is the process of accepting serial changes. The bot will only consider and merge one change at a time. It is forced to do this, because git will only merge one branch at a time.

We could say that the bot operates git, we are fighting over how to influence the bot's activities, and the bot is maintaining the rules of engagement.

[...]

PR #329, opened 2017-05-28
Why Chaos?

collinalexbell:


amoffat:
Love this creativity.  You should definitely think about adding to the [wiki](https://github.com/chaosbot/Chaos/wiki) as well, @SlightlyCyborg, with or without this getting merged in.  Also, come chat with @eukaryote31 and @anythingbot ([relevant comment](https://github.com/chaosbot/Chaos/issues/238#issuecomment-304104976)) in our [chat](https://gitter.im/chaosthebot/Lobby), as they both have expressed similar philosophical ideas about Chaos

PR #331, opened 2017-05-28
Feature: No failed Travis-CI builds

while-loop:
If a pull request failed to pass a Travis-CI build, or tests, then do not treat it as a ready PR.

This PR _should_ reduce the amount broken merged chaos builds, syntax errors, and changes that break unit tests.

Tested on a [non-breaking PR](https://github.com/while-loop/Chaos/pull/16) and a [breaking PR](https://github.com/while-loop/Chaos/pull/15)

### DIFF
--- a/github_api/prs.py
+++ b/github_api/prs.py
@@ -135,12 +135,21 @@ def get_pr_comments(api, urn, pr_num):
 
 
 def has_build_passed(api, statuses_url):
+    """
+        Check if a Pull request has passed Travis CI builds
+    :param api: github api instance
+    :param statuses_url: full url to the github commit statuses.
+           Given in pr["statuses_url"]
+    :return: true if the commit passed travis build, false if failed or still pending
+    """
     statuses_path = statuses_url.replace(api.BASE_URL, "")
 
     statuses = api("get", statuses_path)
 
     if statuses:
         for status in statuses:
+            # check state is success and description of status
+            # the state can be success for Chaosbot statuses, so we double-check if it a Travis CI
             if (status["state"] == "success") and \
                (status["description"] == "The Travis CI build passed"):
                 return True
--- a/github_api/prs.py
+++ b/github_api/prs.py
@@ -1,11 +1,13 @@
-import arrow
 import math
-import settings
+
+import arrow
 from requests import HTTPError
-from . import misc
-from . import voting
+
+import settings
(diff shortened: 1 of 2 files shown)


phil-r:
🎉 This is awesome! Thanks @while-loop

github_api/prs.py
```diff
@@ -132,8 +134,30 @@ def get_pr_comments(api, urn, pr_num):
...
+    statuses_path = statuses_url.replace(api.BASE_URL, "")
+
+    statuses = api("get", statuses_path)
+
+    if statuses:
+        for status in statuses:
+            # check state is success and description of status
+            # the state can be success for Chaosbot statuses, so we double-check if it a Travis CI
+            if (status["state"] == "success") and \
+               (status["description"] == "The Travis CI build passed"):
```
jeroeness, line 154: Is it safe to check if Travis CI build passed this way? If they suddenly decide to change their description, we get locked out.

phil-r: @jeroeness You're right, I think it's better ti check `context` [(Example)](https://developer.github.com/v3/repos/statuses/#example)

while-loop: You guys have better eyes than me 🤦

Fixed #335

amoffat:
This had to be disabled by https://github.com/chaosbot/Chaos/commit/295545f87bf8fe9f5dfc56597dcb53b7e223ebd9

What we were seeing in the chat was that some PRs, like #344 were not getting travis-ci build statuses, and so the vote update statuses and merging/processing was stopping completely because chaos was considering those PRs as failing their travis builds

There was another layer to this as well, and that is that the status url chaos was using, for example [344's status url](https://api.github.com/repos/chaosbot/Chaos/statuses/cea1616848f9182b82a0354feceecaa0e1bdc2fb) was sometimes only ever listing chaos statuses, and never travis statuses.  @phil-r found the [ref status url](https://api.github.com/repos/chaosbot/Chaos/commits/cea1616848f9182b82a0354feceecaa0e1bdc2fb/status) to be slightly more reliable, but it has to be pulled directly from `pr_data["head"]["sha"]` I believe.  He has more details on exactly how to do that

phil-r:
I guess it's a lesson for all of us - we can't blindly rely on third-party service and should make code more failsafe.

PR #332, opened 2017-05-28
Fix website.

Alexays:
- Correct indenting files.
- Fix overflow Ribbon.
- Center box (yes current it's not).
- Fix colors first frame canvas x)
- Correct background canvas which should be fafafa and not ffffff.
edit: Transparent background now.
Add fix for card, add posibility to have a card with big content (ex: voters list).

### DIFF
--- a/server/index.html
+++ b/server/index.html
@@ -29,11 +29,11 @@
         <meta name="msapplication-TileColor" content="#ffffff">
         <meta name="msapplication-TileImage" content="static/assets/icons/ms-icon-144x144.png">
         <meta name="theme-color" content="#ffffff">
-        <meta property="og:url"                content="http://chaosthebot.com/" />
-        <meta property="og:type"               content="website" />
-        <meta property="og:title"              content="ChaosBot" />
-        <meta property="og:description"        content="A social coding experiment that updates its own code democratically" />
-        <meta property="og:image"              content="https://avatars3.githubusercontent.com/u/28798182?v=3&s=400" />
+        <meta property="og:url" content="http://chaosthebot.com/" />
+        <meta property="og:type" content="website" />
+        <meta property="og:title" content="ChaosBot" />
+        <meta property="og:description" content="A social coding experiment that updates its own code democratically" />
+        <meta property="og:image" content="https://avatars3.githubusercontent.com/u/28798182?v=3&s=400" />
         <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/bulma/0.4.1/css/bulma.min.css" />
         <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/4.7.0/css/font-awesome.min.css">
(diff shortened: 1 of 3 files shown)


server/index.html
```diff
@@ -1,65 +1,54 @@
...
+        <link rel="apple-touch-icon" sizes="180x180" href="static/assets/icons/apple-icon-180x180.png">
+        <link rel="icon" type="image/png" sizes="192x192"  href="static/assets/icons/android-icon-192x192.png">
+        <link rel="icon" type="image/png" sizes="32x32" href="static/assets/icons/favicon-32x32.png">
+        <link rel="icon" type="image/png" sizes="96x96" href="static/assets/icons/favicon-96x96.png">
+        <link rel="icon" type="image/png" sizes="16x16" href="static/assets/icons/favicon-16x16.png">
+        <link rel="manifest" href="manifest.json">
+        <meta name="msapplication-TileColor" content="#ffffff">
+        <meta name="msapplication-TileImage" content="static/assets/icons/ms-icon-144x144.png">
+        <meta name="theme-color" content="#ffffff">
+        <meta property="og:url"                content="http://chaosthebot.com/" />
```
aslafy-z: Could be the right time to harmonize the spacings here

Alexays: Yes I will do it ;)

server/static/js/chaos.js
```diff
@@ -4,13 +4,13 @@ var ctx = canvas.getContext("2d");
 canvas.width = 400;
 canvas.height = 200;
 
-ctx.fillStyle = "#000000";
+ctx.fillStyle = "#fffafa";
```
phil-r: is it intended that it's not `#fafafa` ?

server/static/js/chaos.js
```diff
@@ -32,82 +32,84 @@ canvas.onmousemove = function(e) {
...
-    for (x = 0; x < pix_location.length; x++) {
-      var xPos = pix_location[x][0] + 0.1*(char_location[x][0]-pix_location[x][0]);
-      var yPos = pix_location[x][1] + 0.1*(char_location[x][1]-pix_location[x][1]);
-      if (lastX > 0 && lastY > 0) {
-          dx = pix_location[x][0]-lastX;
-          dy = pix_location[x][1]-lastY;
-          if (Math.sqrt(dx*dx+dy*dy) < 10) {
-            yPos = Math.random() > 0.5 ? 0 : canvas.height;
-          }
+  ctx.fillStyle = "#fafafa";
```
phil-r: maybe you can use a variable for `"#fafafa"` ?

server/static/js/chaos.js
Alexays: Yeah thx.

server/static/js/chaos.js
Alexays: It's done :)

phil-r:
Maybe I'm missing something, but why we are not using transparent background for canvas?
Example: https://jsfiddle.net/Phil/1e8Lc757/

PR #333, opened 2017-05-28
1.1x Voting Power for Contributors

MegosAlpha:
Sorry for appearing to be hacking, I guess base64 and zlib was a bad idea. It's my first pull request ever, so that's why it probably looks messy in commit history. All contributors now receive an x1.1 bonus to voting perform, as a way of verifying accounts better.


Alexays:
Before there was also the pr 297 :o

Alexays (commented):
Wtf, why delete voters file ?

server/voters.html
```diff
@@ -1,46 +0,0 @@
...
-                    for (var i = 0; i < list.length; i++) {
-                        tablehtml += "<tr><td>" + list[i].names + "</td><td>" + list[i].votes + "</tr>";
-                    }
-                    tablehtml += "</table>";
-                    result.innerHTML = tablehtml;
-                }
-            }
-        }
-    }
-</script>
```
phil-r: why removing it?

settings.py
```diff
@@ -71,6 +71,21 @@
 # used for calculating how long our voting window is
 TIMEZONE = "US/Pacific"
 
+# contains all members of the galactic federation (everyone against PR#322 at 10:33 MDT)
+# base64 and zlib applied
+FEDERATION = [
+    'phil-r',
```
phil-r: well, I'm kinda proud of myself xD

settings.py
```diff
@@ -71,6 +71,21 @@
 # used for calculating how long our voting window is
 TIMEZONE = "US/Pacific"
 
+# contains all members of the galactic federation (everyone against PR#322 at 10:33 MDT)
+# base64 and zlib applied
```
phil-r: this line is not needed

phil-r:
👎 Although my "name" is in the list, I don't think it's the right way to go. Changing voting power of different people is a tough decision, this can make those people with less voting power less motivated to contribute(except for smittyvb, he is just pain in the ass and deserves it)

MegosAlpha:
I'll lower the power, and thanks for the advice. One sec...

server/voters.html
MegosAlpha: Accident, now fixed

MegosAlpha:
@Leigende the file is back, I deleted it on accident. Also, voters of PR#297 have been added.

yet-another-account:
I'd be almost okay with this if it was an insignificant advantage, like a voting power of 1.05 or something, but even then, I don't really like this idea.

phil-r:
@eukaryote31 I had the same thoughts, Maybe all voters with `contributor` tag should have a bit more voting power(x1.05 like you said)? But I'm against hardcoded `federation` list for sure, sorry @MegosAlpha  😢

MegosAlpha:
@phil-r Ok, I'll work on that. Thanks.

o3LL:
How about making people doing "bad PR" getting less power instead of making an "Elite"?

settings.py
rudehn: You're the leader of the federation :D

MegosAlpha:
I don't quite know how to check if a PR is really bad, because they always get revised. I think that if people do a good PR making them contributors, then it is better because this confirms them as a member of the Chaos community.

MegosAlpha:
@phil-r @eukaryote31 I listened to your advice. It is now a 1.1x bonus to all contributors.

PR #335, opened 2017-05-28
Fix: Change Travis-CI detection check

while-loop:
This new method is less error prone if Travis ever changes the success string. Instead, we check to see if the context of the status is Travis PR related.

As [reviewed](https://github.com/chaosbot/Chaos/pull/331#discussion-diff-118841694R154) by @jeroeness and @phil-r

### DIFF
--- a/tests/github_api/prs_test.py
+++ b/tests/github_api/prs_test.py
@@ -19,31 +19,42 @@ class TestPRMethods(unittest.TestCase):
 
         self.assertTrue(prs.has_build_passed(api, url))
 
-    def test_statuses_returns_failed_travis_build(self):
-        statuses = [{"state": "error",
-                    "context": "continuous-integration/travis-ci/pr"}]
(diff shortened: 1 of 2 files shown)


phil-r:
Oops #336 haha

phil-r:
@while-loop can you grab some extra tests, that I've added?

tests/github_api/prs_test.py
```diff
@@ -35,7 +35,7 @@ def __call__(m, method, path, **kwargs):
         self.assertFalse(prs.has_build_passed(api, url))
 
         statuses = [{"state": "pending",
-                    "description": "The Travis CI build is in progress"}]
+                    "context": "some_other_context"}]
```
phil-r: I think there should be a separate test for "wrong" context

phil-r (approved):

while-loop:
@phil-r I've added your changes

phil-r:
Nice ;)

phil-r:
@while-loop small followup: #337

PR #336, opened 2017-05-28
Check PR status `context` instead of `description`

phil-r:
- added more tests


phil-r:
Closed in favor of #335

PR #338, opened 2017-05-28
Make @smittyvb votes worth about (e^pi-pi)/40

yet-another-account:
Which is about 0.49997749948. Just for the lols. 

Seriously, though, it's a joke. Although, it would be funny to be ~0.000022 away from the voting threshold.


yet-another-account:
@smittyvb would you upvote if I made it 1-(e^pi-pi)/40? That would make your voting power 0.50002250052, or about 0.000023 MORE than you currently have right now! Win-win, amirite?

syvb:
Yep.

yet-another-account:
There you go!

o3LL:
Unfair 😂

PR #339, opened 2017-05-28
[WIP][FIX] Run commands via issue comments

phil-r:
@rudehn I don't know what went wrong with #284 but feel free to use this to fix it


rudehn:
Might as well just use this one. I'm afraid my git-fu isn't good enough to figure out how to use this

PR #340, opened 2017-05-28
[Fixed] Send commands via issues

rudehn:
Alright I'm not sure what happened with the other branch, but here's a clean version.

There's a lot going on, so a brief explanation is as follows.

This provides the functionality to send commands to chaosbot by commenting with a command (see below)  on any given issue. Similar to how voting for PRs goes, there's a window in which votes are gathered. As it is currently set, any negative vote will cause the command to not be run.

Current commands include:
- [x] reopening issues
- [x] closing issues

Suggestions for new commands are welcome and can be implemented in the future.

Thanks go out to @phil-r  for providing a clean version.


rudehn:
I was wondering why the weighted total is 2.5. Then I saw @smittyvb voted :laughing:

rudehn:
aww :(

PR #343, opened 2017-05-28
More browser support for voters.html

syvb:
Before, this code used Object.values, which is only supported in Chrome and Firefox. This new code works replaces the line by mapping the keys array to the variable with all the data, to get the values for each key, causing the same effect. This is supported in almost all modern browsers, and IE9+.


PlasmaPower:
*cautiously upvotes*

yet-another-account:
\<sarcasm>_Any browser with IE in the name cannot be considered modern_\</sarcasm>

PlasmaPower:
@Smittyvb I think the dynamic voting window would've saved you. If not, you could always just add WIP to the title until you get enough votes.

server/voters.html
```diff
@@ -21,7 +21,7 @@
                     var values = keys.map(function(key) {
                         return json[key];
                     });
-
+
```
PlasmaPower: Pro tip: use `--allow-empty` to create an empty git commit.

Edit: also, no trailing whitespace please.

PR #344, opened 2017-05-28
Feature: Give Chaos collaborators more weight

while-loop:
Give chaosbot/Chaos collaborators a bump in weight of 1.5

### DIFF
--- a/github_api/voting.py
+++ b/github_api/voting.py
@@ -105,7 +105,7 @@ def get_pr_review_votes(api, urn, pr_num):
             yield user, vote
 
 
-def get_vote_weight(api, username):
+def get_vote_weight(api, username, is_collaborator):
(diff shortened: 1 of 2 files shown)


github_api/voting.py
```diff
@@ -117,19 +117,35 @@ def get_vote_weight(api, username):
...
+        weight *= 1.5
+
+    # sorry smitty, No food for you (http://e.lvme.me/7mb4jy9.jpg)
     if username.lower() == "smittyvb":
         weight /= 2
 
     return weight
 
 
+def get_collaborators(api, urn):
```
amoffat, line 131: this might be a good function to cache.  see https://github.com/chaosbot/Chaos/blob/master/patch.py#L29-L32

PR #346, opened 2017-05-28
Feature: Chaotic Vote Multiplier

MegosAlpha:
A new take on Wiki Voting Feature Request 1 - Multiplies all users votes by a random number in the (0.5, 1.5] interval. Also adds some simple docstrings.


PlasmaPower:
:-1: the random value changes each time the vote is tallied (which happens a lot)

syvb:
:-1: This would be evaluated every time the votes are calculated, which is every 30 seconds/twice a minute, so this really just makes the threshold lower. This would work if the multiplier was stored somewhere, such as the MySQL database.

mark-i-m:
Lol, we asked for chaos right?

PR #348, opened 2017-05-29
1 Week and 1 Repo Voting Requirements

MegosAlpha:
1 Month means that not much of the community gets to join the project. Not many users are spam accounts, so we may as well increase the number of new users then restrict it to older users only.

### DIFF
--- a/settings.py
+++ b/settings.py
@@ -52,7 +52,7 @@ AFTER_HOURS_START = 22
 AFTER_HOURS_END = 10
 
 # how old do voters have to be for their vote to count?
-MIN_VOTER_AGE = 1 * 30 * 24 * 60 * 60  # 1 month
+MIN_VOTER_AGE = 12 * 60 * 60  # 1 day (12 hrs)
 
 # for a pr to be merged, the vote total must have at least this fraction of the
 # number of watchers in order to pass.  this is to prevent early manipulation of
--- a/settings.py
+++ b/settings.py
@@ -52,7 +52,7 @@ AFTER_HOURS_START = 22
 AFTER_HOURS_END = 10
 
 # how old do voters have to be for their vote to count?
-MIN_VOTER_AGE = 12 * 60 * 60  # 1 day (12 hrs)
+MIN_VOTER_AGE = 7 * 24 * 60 * 60  # 1 week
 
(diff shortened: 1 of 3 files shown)


settings.py
```diff
@@ -52,7 +52,7 @@
 AFTER_HOURS_END = 10
 
 # how old do voters have to be for their vote to count?
-MIN_VOTER_AGE = 1 * 30 * 24 * 60 * 60  # 1 month
+MIN_VOTER_AGE = 12 * 60 * 60  # 1 day (12 hrs)
```
phil-r: soooo 1 day or 12 hours?

MegosAlpha: @phil-r 1 day, not 1 day and 1 night. I just wanted to be clear.

syvb:
👍 1 week sounds good; this project hasn't been around for more than a week, so spam wouldn't be an issue.
~~12 hours is too little time. I could create 50 accounts right now, and in the morning I could spam any PR I wanted with up/downvotes.~~

settings.py
phil-r: there were 24 hours in the day last time I checked...

MegosAlpha: @phil-r Yeah, I see how that would be confusing. It's 1 week now, so no worries.

rhengles:
We need to filter serious users from spam accounts. These users need to be checked for activity somehow, be It repos, commits, or something. They must absolutely not be empty accounts.

MegosAlpha:
@rhengles User now must have at least one public repo before voting.

PlasmaPower:
@MegosAlpha that's easy to create

rhengles:
Yes... I don't think there's an algorithm that can reliably tell users from spam accounts apart. I'd approve a manual whitelist of new users to be added through votes in PRs.

phil-r:
Not sure if someone plans an attack or it's actually important 🤔

rudehn:
So why are PRs being marked as having all checks passed even when they don't have the required number of votes?

PlasmaPower:
@rudehn Chaosbot is down, so its status isn't showing up. That leaves only Travis, which is passing.

rudehn:
ahh thanks! Makes perfect sense

PR #349, opened 2017-05-29
Lower threshold per watchers from 0.05 to 0.04

PlasmaPower:
Right now it's at `6.5`, this would lower it to `5.2`.

### DIFF
--- a/settings.py
+++ b/settings.py
@@ -57,7 +57,7 @@ MIN_VOTER_AGE = 1 * 30 * 24 * 60 * 60  # 1 month
 # for a pr to be merged, the vote total must have at least this fraction of the
 # number of watchers in order to pass.  this is to prevent early manipulation of
 # the project by requiring some basic consensus.
-MIN_VOTE_WATCHERS = 0.05
+MIN_VOTE_WATCHERS = 0.04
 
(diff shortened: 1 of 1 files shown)


phil-r:
👍 Yeah it's a bit too high

rhengles:
oh god no

PlasmaPower:
@rhengles Why not? It's a slight change, and it feels like barely anything gets passed now.

rudehn:
It's also the weekend. That might have something to do with it

PlasmaPower:
@rudehn Good point.

mark-i-m:
I'm not sure that changing the threshold value is the right solution. The threshold will keep needing to be being update as more or fewer people watch and depending on what day of the week. I would rather do something dynamic. For example, you must have more than half the votes of people who have historically voted on that day or some running average or something... These are just ideas, I don't know if any would work...

hongaar:
Yeah.. those were the days:

image

Maybe we can use a moving (weighted) average (i.e. over last week, with more recent days having a heavier weight) of [commit activity](https://developer.github.com/v3/repos/statistics/#get-the-last-year-of-commit-activity-data) or [unique visitors](https://developer.github.com/v3/repos/traffic/#views)?

PR #337, opened 2017-05-28
Update `get_ready_prs` function

phil-r:
- makes it more "flat" 
- more tests! 🎉 
- adds `server/voters.json` to `.gitignore`

PR #350, opened 2017-05-29
Send chaosbot commands via issue comments

rudehn:
Alright I'm not sure what happened with the other branch, but here's a clean version.

There's a lot going on, so a brief explanation is as follows.

This provides the functionality to send commands to chaosbot by commenting with a command (see below)  on any given issue. Similar to how voting for PRs goes, there's a window in which votes are gathered. As it is currently set, any negative vote will cause the command to not be run.

Current commands include:
- [x] reopening issues
- [x] closing issues

Suggestions for new commands are welcome and can be implemented in the future.

Thanks go out to @phil-r  for providing a clean version.

Reopening this since it had full support last time, and just missed 1 vote. Tagging WIP until it reaches a passing state


PlasmaPower:
> chaosbot — remaining: -1:11, vote: 6-0, weighted total: 5.5, threshold: 6.5

You alright there chaosbot?

rudehn:
This might be a bug. I just repulled from the same branch. Also, any idea what's going on with Travis? It was working fine before and I can't seem to find the error it mentioned.

PlasmaPower:
@rudehn Chaosbot requiring Travis didn't work, so it was disabled with a hotfix. Ironically, the hotfix broke Travis by adding an unused variable. That'll be fixed with #337.

mark-i-m:
@rudehn if the record file gets lost, we could see some weird bugs. Have you thought of committing it to the repo? Obviously, this introduces a lot of complexity. Just a suggestion though... I like the PR :+1:

rudehn:
@mark-i-m Yes it will rerun all commands if the file ends up missing. We also could save a datetime and only run commands that have been submitted after that datetime. When we update the datetime it would prevent us from ever retrieving old commands.

cron/poll_read_issue_comments.py
```diff
@@ -0,0 +1,144 @@
...
+
+    # Already ran this command
+    for ids_ran in json_data["comment_ids_ran"]:
+        if comment_id == ids_ran["comment_id"]:
+            __log.debug("Already ran commmand")
+            return False
+
+    # Voting window has passed
+    now = arrow.utcnow()
+    voting_window = gh.voting.get_voting_window(now)
```
hongaar, line 54: Has been removed. I think it can be replace with: `voting_window = gh.voting.get_initial_voting_window(now)`?

hongaar: see #357

PR #352, opened 2017-05-29
api request limit fixes

amoffat:
makes our cooldown a little more effective.  also reenables caching of api requests that were chewing through our limit.

### DIFF
--- a/github_api/__init__.py
+++ b/github_api/__init__.py
@@ -4,6 +4,7 @@ import math
 import requests
 from requests.auth import HTTPBasicAuth
 import logging
+import settings
 
 log = logging.getLogger("github_api")
 
@@ -22,7 +23,13 @@ def compute_api_cooldown(remaining, reset):
     # also account for a potential divide by zero
     actual_remaining = max(remaining - 30, 1)
 
-    return ((reset / actual_remaining) ** 3) / 10.0
+    cooldown = ((reset / actual_remaining) ** 3) / 10.0
+
+    # it doesn't make sense to ever sleep longer than the point where github
+    # refreshes our api counter, but let's also pad that value a little bit
+    cooldown = min(reset + settings.API_COOLDOWN_RESET_PADDING, cooldown)
+
+    return cooldown
 
 
 class API(object):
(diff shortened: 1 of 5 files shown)


PlasmaPower:
Was this causing the broken chaosbot behavior?

Edit: yep

phil-r:
`./chaos.py:17:1: F401 'patch' imported but unused`

PlasmaPower:
You can fix it with `# noqa: F401`

github_api/__init__.py
```diff
@@ -14,7 +14,15 @@ def compute_api_cooldown(remaining, reset):
...
-    return ((reset / remaining) ** 3) / 10.0
+
+    # i've seen github's api fail with a 403 when we had 5 requests remaining,
+    # as confirmed by their response headers.  so we'll take that kind of
+    # fluctuation into account.
+    #
+    # also account for a potential divide by zero
+    actual_remaining = max(remaining - 30, 1)
+
+    return ((reset / actual_remaining) ** 3) / 10.0
```
PlasmaPower: Should we design this function to max out at `reset` when `actual_remaining == 1`?

amoffat: that's a great idea, let me do that

PlasmaPower: You might want to add a bit to reset just in case.

PlasmaPower:
> make travis happy for now

Not quite - two spaces before an inline comment. Remember that you can run flake8 locally.

amoffat:
Thanks guys :+1:

github_api/prs.py
```diff
@@ -182,7 +182,7 @@ def get_ready_prs(api, urn, window):
         # existing on the PRs anymore (somehow..still unsolved), and then PRs
         # were not being processed or updated.  do not use this variable in the
         # if-condition that follow it until that has been solved
-        build_passed = has_build_passed(api, pr["statuses_url"])
+        build_passed = has_build_passed(api, pr["statuses_url"])  # noqa: F841
```
PlasmaPower: Conflict here

PR #356, opened 2017-05-29
Accepts only PR by username starting with "chaos"

betrisey:
To make a pull request, we will have to add `chaos` in front of our username


rudehn:
Some people have the same username for their work account. We can't just change it.

PR #360, opened 2017-05-29
[WIP] Zenofchaos propose

eamanu:
This is a proposal to have a Zen of Chaos.
We can complete the Zen with the principles that we consider important

### DIFF
--- a/ZENOFCHAOS.md
+++ b/ZENOFCHAOS.md
@@ -1,0 +1,10 @@
+# Zen of Chaos.
+This is the Zen of Chaos. We pay tribute to the Chaos. The Chaos join us to accomplish its purpose.
+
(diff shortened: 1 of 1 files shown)


davidak:
I don't see myself represented in these principles.

I'm also not sure if we need a Zen of Chaos at all. Maybe if someone comes with some inspiring or funny principles. Suggestions can be made in the comments.

davidak:
- The chaos inspire us to get new ideas and explore new things.

eamanu:
I not pretend make 'Chaos' my religion (hehe), but I thought that maybe fun

phil-r:
`Religion` is a very strong word... Chaos is a fun and interesting project, but it's not a religion and I hope it will never be it.

PR #361, opened 2017-05-29
PREVENT TAKE-OVERS WITH THIS ONE WEIRD TRICK (SMITTYVB HATES HIM)

PlasmaPower:
This PR adds an autocracy, as specified in [`data/autocracy.txt`](https://github.com/PlasmaPower/chaos/blob/autocracy/data/autocracy.txt). I've hand-picked the members. If you think you should be included, please leave a comment with an explanation of why. The list will be expanded in the future of course.

At least one positive non-author autocracy review on the most recent commit of a PR is required for that PR to be accepted.

**This does not give members of the autocracy much power.** As the autocracy has a wide range of opinions, and only one of them is required to approve the PR, the only thing this prevents is things universally seen as bad, like take-overs.

### DIFF
--- a/github_api/prs.py
+++ b/github_api/prs.py
@@ -12,7 +12,7 @@ from . import voting
 TRAVIS_CI_CONTEXT = "continuous-integration/travis-ci"
 
 
-def merge_pr(api, urn, pr, votes, total, threshold):
+def merge_pr(api, urn, pr, votes, total, threshold, autocracy_satisfied):
     """ merge a pull request, if possible, and use a nice detailed merge commit
     message """
 
@@ -26,7 +26,7 @@ def merge_pr(api, urn, pr, votes, total, threshold):
     if record:
         record = "Vote record:\n" + record
 
-    votes_summary = formatted_votes_summary(votes, total, threshold)
+    votes_summary = formatted_votes_summary(votes, total, threshold, autocracy_satisfied)
 
     pr_url = "https://github.com/{urn}/pull/{pr}".format(urn=urn, pr=pr_num)
 
@@ -79,21 +79,28 @@ Description:
             raise
 
 
-def formatted_votes_summary(votes, total, threshold):
+def formatted_votes_summary(votes, total, threshold, autocracy_satisfied):
     vfor = sum(v for v in votes.values() if v > 0)
     vagainst = abs(sum(v for v in votes.values() if v < 0))
+    autocracy_str = "a" if autocracy_satisfied else "**NO**"
 
-    return ("with a vote of {vfor} for and {vagainst} against, with a weighted total \
-            of {total:.1f} and a threshold of {threshold:.1f}"
-            .strip().format(vfor=vfor, vagainst=vagainst, total=total, threshold=threshold))
+    return """
+with a vote of {vfor} for and {vagainst} against, a weighted total of {total:.1f} \
+and a threshold of {threshold:.1f}, and {autocracy} current autocracy review
+    """.strip().format(vfor=vfor, vagainst=vagainst, total=total, threshold=threshold,
+                       autocracy=autocracy_str)
 
 
-def formatted_votes_short_summary(votes, total, threshold):
+def formatted_votes_short_summary(votes, total, threshold, autocracy_satisfied):
     vfor = sum(v for v in votes.values() if v > 0)
     vagainst = abs(sum(v for v in votes.values() if v < 0))
+    autocracy_str = "✓" if autocracy_satisfied else "✗"
 
-    return "vote: {vfor}-{vagainst}, weighted total: {total:.1f}, threshold: {threshold:.1f}" \
-        .strip().format(vfor=vfor, vagainst=vagainst, total=total, threshold=threshold)
+    return """
+vote: {vfor}-{vagainst}, weighted total: {total:.1f}, threshold: {threshold:.1f}, \
+autocracy: {autocracy}
+    """.strip().format(vfor=vfor, vagainst=vagainst, total=total, threshold=threshold,
+                       autocracy=autocracy_str)
 
 
 def label_pr(api, urn, pr_num, labels):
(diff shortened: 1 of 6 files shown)


rhengles:
I think I should be added mainly because I'm the leader of the voters ranking. Also, I do not try to approve PRs for the lolz, but only ones I think genuinely are in the best interest of the project. I'm planning on contributing something, but Python is not my native language 😛

github_api/prs.py
```diff
@@ -79,21 +79,28 @@ def merge_pr(api, urn, pr, votes, total, threshold):
             raise
 
 
-def formatted_votes_summary(votes, total, threshold):
+def formatted_votes_summary(votes, total, threshold, autocracy_satisfied):
     vfor = sum(v for v in votes.values() if v > 0)
     vagainst = abs(sum(v for v in votes.values() if v < 0))
+    autocracy_str = "an" if autocracy_satisfied else "**NO**"
```
mark-i-m: Should this `"a"` instead of `"an"`

PlasmaPower: Good catch - I previously left out "current" so it read "an autocracy review"

PlasmaPower:
@rhengles Fair enough, added.

PlasmaPower:
People downvoting this PR: would you mind leaving a comment as to why?

mark-i-m:
:-1: (for now)

@PlasmaPower I think the requirement should be at least 2 people from the list. It's easy for one person to miss something, but two are harder to trick... of course, then it's hard to see where the boundary is. Why not 3 or 4 people?

Also, I think that reviewers should have a way of indicating autocracy-review approval separately from personal approval. i.e. I might not personally agree with the PR, but if other people out-vote me, I still want to indicate that it passed autocracy review.

davidak:
I'm against it because i'm not included, so :-1: 

But seriously i think a democracy is important to hear everyones opinion and give him some power to define the way to our all future.

Similar like autocracy, but it isn't excluding anyone.

I'm contributing to open source (and free software) because i think it makes a difference. Excluding me with  things like this PR wastes potential. And including me in your autocracy wouldn't make it better. You could always kick people that don't vote as you want. A little like our representative democracy in germany.

warchildmd:
:-1:
That is not fun

PlasmaPower:
@mark-i-m I think 1 is the baseline, we should start there. It will at least prevent obvious takeover amends. We can move it up if stuff still gets through. At the very least, 1 is better than 0.

If I moved reviews to before votes in the vote tallying, would that be better? That way you could approve changes, but still vote them down. I could also make reviews entirely separate from voting.

LosDrakakos:
Hi !
This is a strictly personnal (maybe not very rational and a bit selfish) opinion but,
I think that 9 members are not chaotic enough for chaotic decision making, even if it's a good mecanism to prevent takeovers.

Aren't takeovers a part of chaos ?

Plus I just discovered Chaosbot and I'd like to be involved in this (but I'm not a developper and I don't know anything about python for now.) so i'm not really happy to see my vote taken away so fast :/

If I may I'd like to share an alternative option, that is :
Boost the vote power of a few hand-picked people and and to automatically add more vote power to people on the top of voters list.

PlasmaPower:
@davidak @cthulhuely My intent isn't to take away anyone's voice. The problem I'm trying to solve here is last minute changes to PRs or hidden changes that people miss when voting for.

> Aren't takeovers a part of chaos ?

We get to choose what is part of chaos and what is not. We get to choose if we even want chaos. This PR aims to reduce chaos. I don't think most people want takeovers, as many dictators wouldn't restore democracy as I did. The project would die out without the voice of the people, as that's the foundation of it.

If I expanded the autocracy to the top 10 voters + the top 10 contributors (there might be some overlap there), would that be better?

mark-i-m:
> I could also make reviews entirely separate from voting.

I think that would work...

I also hear the point made by a lot of others... And this PR has the potential to block PRs simply because I'm not available while I'm at work...

An alternative might be to make the threshold a lot higher unless somebody from the whitelist reviews for autocracy.

hongaar:
👍 

I think you could as well call this something like a representative democracy. If this passes, the contributors kind of chose their parliament, while keeping some direct influence.

Important to note that

> At least one positive non-author autocracy **review** on the most recent commit of a PR is required for that PR to be accepted.

So a **reaction** or **comment** is not enough, the parliament must post an actual review for a PR to pass.

PlasmaPower:
> An alternative might be to make the threshold a lot higher unless somebody from the whitelist reviews for autocracy.

Maybe, but it's just as easy to amend a popular PR as it is to amend an unpopular PR.

> this PR has the potential to block PRs simply because I'm not available while I'm at work

I'm trying to collect a wide range of time zones in my aristocracy, but that is a big concern of mine. Maybe make the extended voting window bigger?

PlasmaPower:
Because of the controversy of this PR, I'm probably going to create another PR after this one containing the more major changes suggested (like reviews not counting as votes, which I think is a great idea).

LosDrakakos:
I think that keeping the same vote threshold, but adding to the requirement at least one review from on of the autocrats could be a good alternative that may allow votes to be useful, and may allow to prevent unwanted takeovers.

davidak:
@PlasmaPower i see your good intention here but i'm not sure if this is the right solution.

Could we give some users whose work is approved by the community a veto right?

That could be a specific emoticon that must be placed in an comment with a explanation why.

Every member has only one veto in a specific timeframe, like one month. (can get changed later)

When every member had used their veto it could get restored for everyone earlier.

PlasmaPower:
@cthulhuely That's exactly what this PR does!

@davidak Seems to me like that's more power. If the goal is to prevent sneaky take-overs, I think at least one approved review is a lot better than no vetos.

mark-i-m:
@cthulhuely 

> I think that keeping the same vote threshold, but adding to the requirement at least one review from on of the autocrats could be a good alternative that may allow votes to be useful, and may allow to prevent unwanted takeovers.

Isn't that the original suggestion?

mark-i-m:
@davidak 

> Every member has only one veto in a specific timeframe, like one month. (can get changed later)

By "member" you mean everyone on the list? I think that could work. But somebody could conceivably still take over with careful timing... IIUC your key observation is that most take-over-PRs would be rejected by popular vote anyway, right?

davidak:
I have no time to think about this further because i have to work. The sad state of political activity in our society.

davidak:
@mark-i-m members are usernames in a textfile, yes.

>your key observation is that most take-over-PRs would be rejected by popular vote anyway, right?

yes, a healthy democracy should work this way. but there are some manipulative "politicians" like @PlasmaPower who tells they will do something that sounds good and later it will be worse for everyone. (he ended democracy and became a dictator earlier)

We need a cricital mass of people who has the time to participate in democratic processes and a way to prevent "take-overs" or similar bad things.

I think with that solution everyones vote should have the same power. But there could be a different thing like veto that don't touches the voting power.

We might implement it like @PlasmaPower suggested here but change it later if we have a better solution.

PlasmaPower:
> he ended democracy and became a dictator earlier

It should be noted that I didn't to anything under my reign and purposely restored democracy ASAP (except I accidentally rejected a PR - I thought my vote there would be normal, but all votes are processed before the server is restarted). If I wanted power I'd already have it.

PlasmaPower:
@davidak Also, excited for the next season of House of Cards coming out tomorrow? I know I am. :wink:

phil-r:
What if based on reviews by "veterans" (parliament or whatever) following labels would be added: 
- **`approved`** (1 or or more `+1` from veterans)
- **`controversial`** (1 or or more `+1` and 1 or or more `-1` from veterans)
- **`disapproved`** (1 or or more `-1` from veterans)
- **`not checked`** (no votes from veterans)

This will leave everyone *equal*, but will allow to quickly see general attitude of people who participated in the project a lot.

PlasmaPower:
@phil-r I still don't think that solves the main problem. #318 could amend dictatorship and it would easily get passed, simply because most people wouldn't notice or wouldn't have time to change their votes.

mark-i-m:
Drat! You have foiled my dastardly scheming :rage:

PlasmaPower:
@mark-i-m no we're working together to erase everyone's code, remember? :wink:

phil-r:
*pulled upstream*

PlasmaPower:
I like the titles of the open PRs right now:

- !!!SEIZE POWER HIJACK AMEND AUTOCRACY PR!!!
- MAKE @SMITTYVB LORD CHANCELLOR FOREVER!
- PREVENT TAKE-OVERS WITH THIS ONE WEIRD TRICK (SMITTYVB HATES HIM)

mark-i-m:
Lol, democracy has discovered advertising :laughing:

PlasmaPower:
@mark-i-m inb4 "partner™ with™ Speedy™ for™ a™ Fast™ VPS™"

settings.py
```diff
@@ -82,3 +82,7 @@
 PR_STALE_HOURS = 36
 
 API_COOLDOWN_RESET_PADDING = 30
+
+# list of autocracy members
+with open("data/autocracy.txt", "r") as f:
```
phil-r, line 87: Maybe it's a common practice and i don't know about it, but what's the point of using file for this? why not a list in `settings.py` for example?

PlasmaPower: IDK, keeps it a bit cleaner I guess?

mark-i-m: I kind of like the file for these long lists. We do the same for lists of emojis for example...

PlasmaPower: It's going to get removed in the next iteration of the PR anyway (where it's a meritocracy).

PlasmaPower:
Okay, closing to setup the meritocracy.

PR #362, opened 2017-05-29
Refactored dynamic_voting_window()

hongaar:
[Model on Google Drive](https://docs.google.com/spreadsheets/d/1qNMzaWe_cSjncqgm3HKy5UIb8LdhNgUm916OXZqO_4s/edit?usp=sharing)

Not perfect, but at least we use `log()` now ;)

### DIFF
--- a/github_api/misc.py
+++ b/github_api/misc.py
@@ -1,3 +1,4 @@
+from math import log
 
 def dt_to_github_dt(dt):
     return dt.isoformat() + "Z"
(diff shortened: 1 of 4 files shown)


github_api/misc.py
```diff
@@ -1,3 +1,4 @@
+from math import log
```
mdcfe, line 2: There needs to be another line here for Travis not to fail.

phil-r:
why everyone loves `log` so much? [Sigmoid functions](https://en.wikipedia.org/wiki/Sigmoid_function) are way cooler imho 🤓

hongaar:
@phil-r looks better suited for the job indeed. gotta run now but feel free to edit my fork.

phil-r:
@hongaar we can modify later, your solution is also cool ;)

mark-i-m:
I like logs. Where else would I look to find out what caused my crash? :stuck_out_tongue:

PR #368, opened 2017-05-29
 Make @smittyvb votes worth about (e^pi-pi)/40 again

sugar700:


sugar700:
"remaining: -4:48"

yay

phil-r:
@xfix I think chaosbot looks at the time of the latest commit not at the time when the PR was created(it was discussed a lot), if you want to resubmit this - just push an empty PR or amend. ;)

PR #369, opened 2017-05-29
You won't believe this PR!!! Commit 3 will amaze you!!

rudehn:
Leave a message with the state of the vote and how long until the voting window is over.

Looks like this

image

There are also some bug fixes that are required for this to work.

### DIFF
--- a/cron/poll_read_issue_comments.py
+++ b/cron/poll_read_issue_comments.py
@@ -91,7 +91,7 @@ def set_time_remaining(api, comment_id, comment_txt):
 
     data = {
         "time_remaining": seconds_remaining,
-        "command": comment_txt # Keep this fresh so nobody edits their command post..
+        "command": comment_txt  # Keep this fresh so nobody edits their command post..
         }
     update_db(comment_id, data, db=command_history)
 
--- a/cron/poll_read_issue_comments.py
+++ b/cron/poll_read_issue_comments.py
@@ -2,6 +2,7 @@ import logging
 import arrow
 import json
 import os
+import re
 from os.path import join, abspath, dirname
 
 import settings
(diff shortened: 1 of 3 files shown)


phil-r:
@rudehn I'm soooo amazed 😆 
But seriously - good job!

mark-i-m:
Pretty soon the titles of our PRs are going to be completely meaning less.

yet-another-account:
@mark-i-m `PR #3731: "b2ggaGkgdGhlcmUgby8=! "`

mark-i-m:
How is this still at 7hrs left?

rudehn:
I think something is broke with the status reporting. I did push to a separate branch about an hour ago. Does that reset the timer?

mark-i-m:
Hmm... I don't know. Could you try pushing some random change now? If the timer resets back to the full window, then we know that either Github or Chaosbot is broken :smile:

rudehn:
Yup it reset.

mark-i-m:
Huh, well now we know... Thanks @rudehn

rudehn:
@mark-i-m 

> def get_pr_last_updated ...
 repo = pr_data["head"]["repo"]

This is the culprit. This is repo specific, not branch specific.

mark-i-m:
Yep... I've been looking through the GitHub API, and I think what we have to do is use `pr_data["head"]["sha"]` to get the commit sha for the HEAD of the merging branch. Then, use https://developer.github.com/v3/git/commits/ to get the commit date/time.

PlasmaPower:
@mark-i-m BAD IDEA! Commit time can be modified with `--date`. The repo push time is the only secure way.

rudehn:
There's also the `head` parameter when searching pull requests.

> Filter pulls by head user and branch name in the format of user:ref-name. Example: github:new-script-format.

mark-i-m:
So then, people just shouldn't work on more than one feature at a time?

mark-i-m:
@rudehn Could you explain how you propose using the `head` parameter more?

mark-i-m:
Attempt 2: It looks like we might be able to use the Events API to get the last PushEvent for a repo... Then filter by the `ref` that matches the PR. But that would be a lot of work...

mark-i-m:
In other words, we would want to do something like this:

```
pr_ref = pr["head"]["ref"]
pr_repo = pr["head"]["repo"]
pr_owner = pr["user"]["login"]

events = GET /repos/{pr_owner}/{pr_repo}/events
events = filter(events, lambda e: e["payload"]["ref"] = pr_ref)
last_updated = max(sorted(map(events, lambda e: e["created_at"])))
```

PR #370, opened 2017-05-29
Voting is now more realistic

mgardne8:
Your vote may or may not count, you don't actually know, Just like in real democracies!!

### DIFF
--- a/github_api/voting.py
+++ b/github_api/voting.py
@@ -1,5 +1,6 @@
 import arrow
 from emoji import demojize
+from random import random
 
 from github_api.misc import dynamic_voting_window
 from . import prs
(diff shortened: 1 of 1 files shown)


PlasmaPower:
:-1: We aim to be better than actual democracies, not to emulate them.

PlasmaPower:
You know a PR is unpopular when the author upvotes it.

Edit: then changes it to :tada: after this comment.

mgardne8:
Your dictatorship holds power no more sir, begone.

PlasmaPower:
@mgardne8 My dictatorship doesn't, by choice. But my vote does, and my voice does.

mgardne8:
MAYBE!

PlasmaPower:
This PR also has the same problem as previous RNG PRs, where the RNG is recalculated every time the vote is tallied, in this case effectively nullifying it.

mgardne8:
Yeah, we have a broken polling machine, you never know what you'll get, you might thing you're winning, get one more vote and then be horribly out of luck.

PlasmaPower:
@mgardne8 Not how it works - do you know about the extended voting window?

PR #371, opened 2017-05-29
/vote close on stale issues

mark-i-m:
- Make chaosbot poll every once in a while to see if older issues have been updated.

- Issues that have not been commented on in 3 days get a `/vote close` from chaosbot

- If you want to keep the issue open, you can down vote on that post via the normal mechanisms

Motivation: clean up the clutter so that only issues that people are actively discussing are on the issue viewer...

### DIFF
--- a/cron/poll_issue_close_stale.py
+++ b/cron/poll_issue_close_stale.py
@@ -24,19 +24,22 @@ def poll_issue_close_stale():
     __log.info("There are currently %d open issues" % len(issues))
 
     for issue in issues:
+        number = issue["number"]
         last_updated = arrow.get(issue["updated_at"])
 
         now = arrow.utcnow()
         delta = (now - last_updated).total_seconds()
 
+        __log.info("Issue %d has not been updated in %d seconds" % (number, delta))
+
         if delta > settings.ISSUE_STALE_THRESHOLD:
-            __log.info("Vote close issue %d" % issue["id"])
+            __log.info("/vote close issue %d" % number)
 
             # leave an explanatory comment
             body = "This issue hasn't been active for a while." + \
                     "To keep it open, react with :-1: on the `vote close` post."
-            gh.comments.leave_comment(api, settings.URN, issue["id"], body)
+            gh.comments.leave_comment(api, settings.URN, number, body)
 
             # then vote to close
             body = "/vote close"
-            gh.comments.leave_comment(api, settings.URN, issue["id"], body)
+            gh.comments.leave_comment(api, settings.URN, number, body)
--- a/cron/poll_issue_close_stale.py
+++ b/cron/poll_issue_close_stale.py
@@ -1,7 +1,5 @@
 import arrow
 import logging
-import os
-from os.path import join, abspath, dirname
 
 import settings
 import github_api as gh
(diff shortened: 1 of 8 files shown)


cron/poll_issue_close_stale.py
```diff
@@ -0,0 +1,45 @@
...
+def poll_issue_close_stale():
+    """
+    Looks through all open issues. For any open issue, if the issue is
+    too old and has not been recently commented on, chaosbot issues a
+    /vote close...
+    """
+
+    __log.info("Checking for stale issues...")
+
+    api = gh.API(settings.GITHUB_USER, settings.GITHUB_SECRET)
```
amoffat: One thing I noticed that may be a potential source of api rate limit bugs.  When a new api instance is created, it sets its remaining requests to infinity, and so assumes that it can proceed with any request and will succeed.  Sometimes this may not succeed if we're low on api requests though.  One solution I can think of is use a single api instance at the top-level cron module and pass it into each cron polling function, or (if it's difficult to pass arguments to be called with the callback) bake it into the function with [functools.partial](https://docs.python.org/3.6/library/functools.html#functools.partial)

settings.py
```diff
@@ -39,6 +39,7 @@
 # the number of seconds chaosbot should sleep between polling for ready prs
 PULL_REQUEST_POLLING_INTERVAL_SECONDS = 30
 ISSUE_COMMENT_POLLING_INTERVAL_SECONDS = 60 * 10  # 10 min window on polling comments
+ISSUE_CLOSE_STALE_INTERVAL_SECONDS = 60 * 10  # 10 min window on polling issues
```
phil-r: I think It's okay to do it once an hour, or even less frequently

cron/poll_issue_close_stale.py
mark-i-m: Ah, I was wondering if there was a catch to using `gh.API` everywhere... Will update and push

settings.py
mark-i-m: @phil-r I agree completely. I wanted to do it every other hour or so. The problem is that if a PR is accepted in the middle of that interval, the timer resets... which means that the cron never runs. :hourglass:

phil-r: 🤔 Is there an option to `run right now` and repeat every n seconds/minutes/hours?

phil-r: can't we just run `poll_issue_close_stale` on startup?

mark-i-m: We could do something like this in cron/__init__.py:
```python
def schedule_jobs(api):
    schedule.every(settings.PULL_REQUEST_POLLING_INTERVAL_SECONDS).seconds.do(
            lambda: poll_pull_requests(api))
    schedule.every(settings.ISSUE_COMMENT_POLLING_INTERVAL_SECONDS).seconds.do(
            lambda: poll_read_issue_comments(api))
    schedule.every(settings.ISSUE_CLOSE_STALE_INTERVAL_SECONDS).seconds.do(
            lambda: poll_issue_close_stale(api))
    # call manually the first time
    poll_issue_close_stale(api)
```

phil-r: :+1: looks fine for me

mark-i-m: @phil-r ^^^ Is that what you meant?

mark-i-m: Sorry, your post didn't show up until now

PR #377, opened 2017-05-29
Add a meritocracy

PlasmaPower:
I decided to not go with a clickbait title this time :smile:

Look at #361 for more details, but **I tried to give the meritocracy minimal power. Democracy still decides the outcome >90% of the time.** This is just to prevent against sneaky take over attempts.

Basis of how it works:

> At least one positive non-author meritocracy review on the most recent commit of a PR is required for that PR to be accepted.

Changes from #361:

- Autogenerate meritocracy list (currently top 10 contributors + top 10 voters)
- Don't count reviews as votes (so members of the meritocracy can allow a PR without voting for it)

### DIFF
--- a/github_api/voting.py
+++ b/github_api/voting.py
@@ -31,6 +31,7 @@ def get_votes(api, urn, pr, meritocracy):
         if (vote > 0 and is_current and vote_owner != pr_owner
                 and vote_owner.lower() in meritocracy):
             meritocracy_satisfied = True
+            break
 
     # by virtue of creating the PR, the owner defaults to a vote of 1
     if votes.get(pr_owner) != -1:
--- a/github_api/voting.py
+++ b/github_api/voting.py
@@ -10,7 +10,7 @@ from . import repos
 import settings
 
 
-def get_votes(api, urn, pr):
+def get_votes(api, urn, pr, meritocracy):
     """ return a mapping of username => -1 or 1 for the votes on the current
     state of a pr.  we consider comments and reactions, but only from users who
     are not the owner of the pr.  we also make sure that the voting
@@ -18,6 +18,7 @@ def get_votes(api, urn, pr):
     can't acquire approval votes, then change the pr """
 
     votes = {}
+    meritocracy_satisfied = False
     pr_owner = pr["user"]["login"]
     pr_num = pr["number"]
 
@@ -26,15 +27,16 @@ def get_votes(api, urn, pr):
         votes[voter] = vote
 
     # get all the pr-review-based votes
-    for vote_owner, vote in get_pr_review_votes(api, urn, pr_num):
-        if vote and vote_owner != pr_owner:
-            votes[vote_owner] = vote
+    for vote_owner, is_current, vote in get_pr_review_votes(api, urn, pr):
+        if (vote > 0 and is_current and vote_owner != pr_owner
+                and vote_owner.lower() in meritocracy):
+            meritocracy_satisfied = True
 
     # by virtue of creating the PR, the owner defaults to a vote of 1
     if votes.get(pr_owner) != -1:
         votes[pr_owner] = 1
 
-    return votes
+    return votes, meritocracy_satisfied
 
 
 def get_pr_comment_votes_all(api, urn, pr_num):
@@ -94,15 +96,16 @@ def get_comment_reaction_votes(api, urn, comment_id):
             yield reaction_owner, vote
 
 
-def get_pr_review_votes(api, urn, pr_num):
+def get_pr_review_votes(api, urn, pr):
     """ votes made through
     https://help.github.com/articles/about-pull-request-reviews/ """
-    for review in prs.get_pr_reviews(api, urn, pr_num):
+    for review in prs.get_pr_reviews(api, urn, pr["number"]):
         state = review["state"]
         if state in ("APPROVED", "DISMISSED"):
             user = review["user"]["login"]
+            is_current = review["commit_id"] == pr["head"]["sha"]
             vote = parse_review_for_vote(state)
-            yield user, vote
+            yield user, is_current, vote
 
 
 def get_vote_weight(api, username):
--- a/README.md
+++ b/README.md
@@ -50,8 +50,7 @@ Wondering how to contribute? Try implementing a feature requested [here](https:/
 Votes on a PR are determined through following mechanism:
 * A comment that contains an emoji signifying a vote somewhere in the body counts as a vote for
   or against the PR.
-* Same for reactions on the PR itself and an accept/reject [pull
-  request review](https://help.github.com/articles/about-pull-request-reviews/)
+* Same for reactions on the PR itself
 * The PR itself counts as :+1: from the owner, unless they vote otherwise.
 * Voting goes on for the duration of the voting window - currently 2 or 3 hours,
   depending on the local server time.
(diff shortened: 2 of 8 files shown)


mark-i-m:
For anyone else looking through this PR, it's a lot easier to read if you append `?w=1` to the URL, which gets rid of whitespace changes (e.g. indentation changes).

PlasmaPower:
@mark-i-m yes definitely do that

*hopes nobody notices my trojan horse coded in Whitespace*

github_api/voting.py
```diff
@@ -26,15 +27,16 @@ def get_votes(api, urn, pr):
         votes[voter] = vote
 
     # get all the pr-review-based votes
-    for vote_owner, vote in get_pr_review_votes(api, urn, pr_num):
-        if vote and vote_owner != pr_owner:
-            votes[vote_owner] = vote
+    for vote_owner, is_current, vote in get_pr_review_votes(api, urn, pr):
+        if (vote > 0 and is_current and vote_owner != pr_owner
```
mark-i-m, line 31: Should this be `vote_owner.lower() != pr_owner.lower()`? i.e. are we sure the cases match?

phil-r:
👍 What can possibly go wrong?

github_api/voting.py
PlasmaPower: That's how it was, and it works since they both come from the GitHub API.

mark-i-m: Ok, that's what I thought

PlasmaPower:
@phil-r your comment keeps duplicating client-side:

screenshot

No idea why. It seems to be a GitHub bug. The bug survives a refresh, although the duplicated comment versions don't.

Edit: posting another comment seems to have fixed it.

phil-r:
@PlasmaPower I've also noticed, refresh helped, maybe I've triggered some easter egg? 🐰 🥚

PlasmaPower:
@phil-r I think it's actually a bug with reviews. It only seems to happen when the comment is next to a review. Overall, reviews are poorly implemented IMO.

PlasmaPower:
When you fix a minor mistake but it resets the voting window :cry:

I guess #378 wins the merge conflict race.

mark-i-m:
Hey! :tada:

mark-i-m:
What was the fix?

PlasmaPower:
@mark-i-m In `patch.py` I cached `gh_api.prs.get_contributors` not `gh_api.repos.get_contributors`.

mark-i-m:
Btw @PlasmaPower you down voted your PR by crying in the post above...

PlasmaPower:
@mark-i-m lol oops. :+1: I think that fixes it?

Edit: yep

PlasmaPower:
Reminds me of when reactions to comments counted for the PR as a whole.

PlasmaPower:
Had to do another slight fix: `.iteritems()` was removed from Python 3. Changed to just `.items()`

mark-i-m:
@PlasmaPower Are you squashing the commits in your own fork?

PlasmaPower:
@mark-i-m Oh yeah, I probably shouldn't now that chaosbot squashes them for us. Doesn't make much of a difference though, and there's no added security - I could always amend a malicious change to a previous commit.

mark-i-m:
I have my :eye: on you

PlasmaPower:
*waits until @mark-i-m goes to sleep* /s

PlasmaPower:
Or do you sleep with that eye open?

mark-i-m:
You don't know if I sleep (at the same time as you)

PlasmaPower:
I just realized I forgot to pass `reverse=True` to sort, so it was an opposite meritocracy... oops! Will be fixed soon, just double checking a few things.

davidak:
@PlasmaPower this solution looks very good!

So at least one member of this group has to do a review seperately from voting. So people needs to know if they are in this group and that it's helpful if they review. Also that they have responsibility for the project doing it.

It would be good if you add that to the readme and we need some kind of news section to inform the community of such important changes.

davidak:
@PlasmaPower If this get's merged you can mention everyone currently in the list so they know that they now have this opportunity.

PlasmaPower:
Current meritocracy: @chaosbot, @muchzer, @pchauncey, @amoffat, @hongaar, @rhengles, @rudehn, @xyproto, @eukaryote31, @ad-m, @mark-i-m, @smittyvb, @reddraggone9, @plasmapower, @phil-r

I've updated the README, pushing the changes in a second. The meritocracy will be logged every time chaosbot checks for new PRs.

(yes I know chaosbot is in the list, but hardcoding chaosbot is both unnecessary and probably a bad idea. chaosbot will also fade from the list as merge commits give way to squashes)

PlasmaPower:
Okay, both of those changes have been pushed. Ready to merge! :tada:

Other slight updates can wait for another PR, I'll only add on important fixes.

(now to wait 9 hours because the extended voting window is broken... :sleeping:)

davidak:
>Current meritocracy

@PlasmaPower shoudn't that be 20 people?

andrewda:
@davidak some people are duplicates – they're in both the top 10 contributors and the top 10 voters

andrewda:
We should eventually have a webpage up similar to http://chaosthebot.com/voters.html to show the current meritocracy. I can work on that tomorrow.

PlasmaPower:
@andrewda I agree, the reason I didn't implement it now is because it'd be the first dynamic part of the server.

PlasmaPower:
Rebased - fixed `settings.py` conflicts.

PlasmaPower:
The Travis failure appears to be from master.

Edit: yep, fixed in #390

andrewda:
@PlasmaPower 
> @andrewda I agree, the reason I didn't implement it now is because it'd be the first dynamic part of the server.

It doesn't have to be dynamic. How I currently have it written is to update a `meritocracy.json` every time it's generated, just like how the `voters.json` works. Both approaches are a little bit gross, but we can work on refactoring that and/or using Django in another PR.

tarunbatra:
This is something nice, @PlasmaPower. Though it chooses the same set of ppl, more or less, but look at the response. 😄

PlasmaPower:
@tarunbatra It's also nice that it automatically updates itself. I might adjust the config later though.

PlasmaPower:
> How I currently have it written is to update a meritocracy.json every time it's generated, just like how the voters.json works.

The difference there is that `voters.json` needs to survive restarts, but `meritocracy.json` doesn't (it'd be essentially write-only). I think a dynamic server is the better solution.

andrewda:
I completely agree

hongaar:
well well well... another regime change by @PlasmaPower!

andrewda:
For some reason, this duplicated the object in [`voters.json`](http://chaosthebot.com/voters.json) and now it's invalid. This breaks the [voters page](http://chaosthebot.com/voters.html).

```json
{ "amoffat": 12468, "g19fanatic": 1442, ... }{ "amoffat": 12468, "g19fanatic": 1442, ... }
```

PlasmaPower:
That might also remove the top 10 voters part of the meritocracy. Anybody know what's causing it?

andrewda:
It's also crashing repeatedly http://chaosthebot.com:8081/chaos-stderr.log

PlasmaPower:
Oh I think that's caused by two PRs being merged at the same time. I'll get a fix ready.

rudehn:
So what happens when somebody who is in the meritocracy reviews the PR, but falls out of the meritocracy before the PR is merged? Does the review no longer count?

andrewda:
I believe that the next time the PR gets polled, the review will no longer count.

rudehn:
Is that working as intended, or a bug?

PlasmaPower:
That's how it works, and there's really no way to fix it unless we get the database up and running (even then it'd be a bit of a pain).

mdcfe:
@PlasmaPower This *is* supposed to be chaos - who knows?

PR #381, opened 2017-05-30
[WIP] Start getting an interface for MySQL

rudehn:
Just starting to flesh out some code

### DIFF
--- a/db/db.py
+++ b/db/db.py
@@ -47,3 +47,4 @@ class DB(object):
 # user = 'chaos'
 # password = 'chaos'
 # db = 'db'
+# comment
--- a/db.py
+++ b/db.py
@@ -4,6 +4,7 @@ import logging
 
 __log = logging.getLogger("DB")
 
+
 class DB(object):
     def __init__(self):
         self.conn = None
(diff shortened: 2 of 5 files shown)


PlasmaPower:
Mark as WIP? or do you think it's done enough to merge as a base?

hongaar:
Would sqlite be an option? Waaay less overhead and waaay more better unless we're going to build the next social network...?

rudehn:
I would prefer it, but mysql was already present in the Dockerfile. I'm not sure if it's being used anywhere in production though.

syvb (approved):

syvb (approved):

db/db.py
```diff
@@ -0,0 +1,50 @@
+import pymysql
+import logging
+
+
```
andrewda, line 4: Remove extra new line (example from elsewhere in the code: https://github.com/chaosbot/Chaos/blob/9dabaf3349f64c9b7a4b9d32e3a1ccb7afa8122e/cron/poll_issue_close_stale.py#L7)

andrewda (changes requested):
Looks good, just one tiny change

rudehn:
See #456

hongaar:
Let's keep it though, who knows we'll be building the next social network anyway: #461

PR #382, opened 2017-05-30
Never change this valuegit diff! EVERgit diff!

mark-i-m:
Lol, That's what I get for including `!!` in a bash command.

Ping @Ayyko... What do you think?

### DIFF
--- a/settings.py
+++ b/settings.py
@@ -33,8 +33,12 @@ GITHUB_USER, GITHUB_REPO = URN.split("/")
 
 HOMEPAGE = "http://chaosthebot.com"
 
-# TEST SETTING PLEASE IGNORE
+# TEST SETTING
 TEST = False
+
+# This exists solely for people to ignore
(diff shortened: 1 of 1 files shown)


settings.py
```diff
@@ -33,9 +33,13 @@
 
 HOMEPAGE = "http://chaosthebot.com"
 
-# TEST SETTING PLEASE IGNORE
+# TEST SETTING
```
PlasmaPower, line 36: The author of this forgot to add `TEST COMMENT PLEASE IGNORE`. Classic mistake.

mark-i-m: No. No. No. You're suppose to pay very close attention to that variable. It's `TSPI` that you should ignore...

PlasmaPower: What `TSPI`? I don't see anything :wink:

mark-i-m: Look at line 41.

PR #375, opened 2017-05-29
🤖 Updated get_vote_sum, and added tests

phil-r:

PR #383, opened 2017-05-30
🤓📈 Use sigmoid function to calculate dynamic_voting_window

phil-r:
- updated tests
- reduced `EXTENDED_VOTE_WINDOW` to 8

[Interactive graphic](https://www.wolframalpha.com/input/?i=plot+8%2F(e%5E(-x%2F12)+%2B+1)+,++x+%3D+-0..60)

Not so interactive:
image

### DIFF
--- a/github_api/misc.py
+++ b/github_api/misc.py
@@ -1,4 +1,4 @@
-from math import log
+from math import exp
 
 
 def dt_to_github_dt(dt):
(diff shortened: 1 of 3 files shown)


phil-r:
@hongaar it's a followup to #362 ^_^

hongaar:
Much more better than log!

hongaar:
We'll need to look into the mechanism which decides to use the extended window, because for PRs such as this one, we should have the default (short) one! Also, I think we can get rid of the 'after hours' window now that we have an extended one..

_edit_ fixed in #375 I think?

PR #376, opened 2017-05-29
Implement voters list as card.

Alexays:
may-30-2017-1-08-am

PR #354, opened 2017-05-29
Position prefix + monospace font on voters.html

mdcfe:
Adds a prefix to entries in the voters table showing how far down they are. Also use a monospace font - which is a start on #319 point 2.

PR #384, opened 2017-05-30
Clean up voters page

andrewda:
This cleans up the voters.html page from #313.

### DIFF
--- a/server/voters.html
+++ b/server/voters.html
@@ -1,59 +1,69 @@
-<center>
-    <h1>Voters on Chaos</h1>
-    <div id="result"></div>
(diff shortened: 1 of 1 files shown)


mdcfe:
I considered this in #354 but decided against it in case anyone else had other plans like replacing that page.

andrewda:
@md678685 Yeah, it should definitely be replaced or put on the homepage at some point, but may as well make it as nice as possible for the time being (or at least conform to some basic HTML styling) 😄

mdcfe:
#376 puts a shortened voters list on the homepage.

PR #385, opened 2017-05-30
Fix keypress listener

mdcfe:
Shhhh I can do listeners (see #367)


mdcfe:
Travis failing due to #371.

PR #388, opened 2017-05-30
Equal voting window # 2

hongaar:
There's now an extended voting window for slow or controversial PRs, thought we might as well skip the _preferential treatment_ of timezones.

Related: #282

### DIFF
--- a/github_api/voting.py
+++ b/github_api/voting.py
@@ -196,20 +196,9 @@ def friendly_voting_record(votes):
     return record
 
 
-def get_initial_voting_window(now):
-    """ returns the current voting window for new PRs.  currently, this biases
-    a smaller window for waking hours around the timezone the chaosbot server is
-    located in (US West Coast) """
-    local = now.to(settings.TIMEZONE)
-    lhour = local.hour
-
-    hours = settings.DEFAULT_VOTE_WINDOW
-    if (lhour >= settings.AFTER_HOURS_START or
-            lhour <= settings.AFTER_HOURS_END):
-        hours = settings.AFTER_HOURS_VOTE_WINDOW
-
-    seconds = hours * 60 * 60
-    return seconds
+def get_initial_voting_window():
+    """ Returns the current voting window for new PRs in seconds. """
+    return settings.DEFAULT_VOTE_WINDOW * 60 * 60
 
 
 def get_extended_voting_window(api, urn):
--- a/cron/poll_pull_requests.py
+++ b/cron/poll_pull_requests.py
@@ -1,4 +1,3 @@
-import arrow
 import logging
 import json
 import os
--- a/cron/poll_pull_requests.py
+++ b/cron/poll_pull_requests.py
@@ -17,8 +17,7 @@ def poll_pull_requests(api):
     __log.info("looking for PRs")
 
     # get voting window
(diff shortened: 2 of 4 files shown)


flibustier:
St. hongaar, holy advocate of equality 😄

PlasmaPower:
:+1: The meritocracy means we don't have to worry about sneaky changes as much, so we can remove the "after hours".

Did you mean to remove `AFTER_HOURS_VOTE_WINDOW` and related settings from `settings.py`? It looks like you didn't touch them.

hongaar:
@PlasmaPower got left behind, fixed now, thanks!

phil-r (approved):

phil-r:
@hongaar I've fixed linting issues, the other one will be fixed in #404

hongaar:
@phil-r awesome, thanks!

btw wonder how GitHub determines who becomes the author of the squashed commit now... :)

PlasmaPower (approved):

andrewda (approved):

phil-r:
@hongaar I think it's the author of the PR, but lets see

PlasmaPower:
My review without a vote was intentional. I'm neutral to this change. Maybe we could do something based on common commit times (you can get that from the GitHub API).

PlasmaPower:
@phil-r It is the author of the PR. Your review was before the last commit though.

PlasmaPower:
There's really no way to enforce this, but I'd prefer we not vote on changes we make.

hongaar:
@PlasmaPower what do you mean?

PlasmaPower:
@hongaar make the voting window shorter in periods where more commits happen.

Edit: if you're talking about my last comment, then: phil-r committed to your branch, so I'd prefer he not vote on this PR after his changes.

phil-r:
@PlasmaPower I don't know how to remove review xD

PlasmaPower:
@phil-r You're fine there - it was for a previous commit so it doesn't count. Once #396 gets merged (38 minutes unless changes happen) you can re-review the commit and the newer review will count.

phil-r:
https://github.com/blog/2265-dismissing-reviews-on-pull-requests Soooo confusing, can't find this in the ui

PlasmaPower:
@phil-r that's for maintainers of the repo, not the author of the review

github_api/voting.py
```diff
@@ -196,20 +196,9 @@ def friendly_voting_record(votes):
...
-    hours = settings.DEFAULT_VOTE_WINDOW
-    if (lhour >= settings.AFTER_HOURS_START or
-            lhour <= settings.AFTER_HOURS_END):
-        hours = settings.AFTER_HOURS_VOTE_WINDOW
-
-    seconds = hours * 60 * 60
-    return seconds
+def get_initial_voting_window():
+    """ Returns the current voting window for new PRs in seconds. """
+    return settings.DEFAULT_VOTE_WINDOW * 60 * 60
```
reddraggone9, line 201: This could set directly in `settings.py` (e.g. `DEFAULT_VOTE_WINDOW_SECONDS`) instead of having a function since it's so simple now. There's no need to recalculate it every time.

reddraggone9 (approved):

PR #390, opened 2017-05-30
Fix flake8 (and therefore Travis)

PlasmaPower:


pchauncey (approved):

PlasmaPower:
@pchauncey Friendly reminder that after #377 passes, reviews no longer also count as votes, so you might want to upvote as well.

pchauncey:
@PlasmaPower Thanks, upvoted

github_api/misc.py
```diff
@@ -1,7 +1,6 @@
 from math import exp
 
 
-
```
flibustier, line 4: His talents as a developer are undeniable. This is by far the most beautiful line of code I've seen in my life!

phil-r: it's basically a lack of line..

PlasmaPower: You know what they say, the fastest code is no code :P

rudehn: We sure saved a lot of time parsing that file now 😆

PR #391, opened 2017-05-30
[WIP] Refactor voters persistence

hongaar:
Votes were always saved, even if PR was not in voting window or some exception occurred. Votes are now only persisted to `voters.json` if, and only if, a PR is either merged or closed.

Current data is corrupted. Should we remove current `voters.json` and start over?

- [x] wait for #393
- [x] wait for #456
- [ ] create migration

### DIFF
--- a/cron/poll_pull_requests.py
+++ b/cron/poll_pull_requests.py
@@ -1,6 +1,5 @@
 import arrow
 import logging
-import json
 import os
 import sys
 from os.path import join, abspath, dirname
@@ -72,6 +71,9 @@ def poll_pull_requests(api):
                 pr_owner = pr["user"]["login"]
                 gh.users.follow_user(api, pr_owner)
 
+                # save votes to voting record
+                gh.voting.persist_votes(votes)
+
                 needs_update = True
 
         else:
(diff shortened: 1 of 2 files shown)


rudehn:
I would delete the file. This does bring up an idea I've been having, namely the idea of some kind of migration/service tool. This tool would run scripts once and only once to do tasks such as moving/deleting files, setting up db tables, etc.

andrewda:
We would definitely need to remove `voters.json` if this is to get merged.

andrewda:
There are currently 318 PRs. How difficult would it be to make some kind of tool which gathers voting data from all previous PRs and add it to the `voters.json` if the file doesn't exist?

hongaar:
@rudehn that's awesome. building upon that, how about we parse a special /depends on #XX command in PR descriptions to indicate a PR can only be merged if another one is merged first.

@andrewda certainly possible. good use case for a migration tool.

I'll mark as WIP for now...

phil-r:
Just make `voters2.json` 😉

flibustier:
Is it possible to use github id instead of username? (I am currently split! 😄)

And maybe it's time for… Database!

hongaar:
@flibustier good suggestions, if the `voters.json` is being reset anyway, let's set it up properly while we're at it. I'm creating a migrations script now, we can use that to initialise a sqlite database?

Also, we can record more detailed records this way, i.e. total amount of votes, total positive & negative votes, and total aligned votes (i.e. votes with same direction as final PR status)

something like:

| column | example |
|---------|-|
| user_id | `int` 4328
| user_last_username | `text` flibustier
| votes | `int` 54 (abs amount of votes)
| positive_votes | `int` 20
| negative_votes | `int` 34
| aligned_votes | `int` 42
| percentage_aligned | `real` 0.7777...
| last_vote | `int` 1496158014 (date of last vote)

@PlasmaPower will solve after #377 is merged

PlasmaPower:
Conflicts with #377

mdcfe:
@hongaar If you're going to use a database, you should probably also back it up every time ChaosBot accepts a new PR in case the PR causes corruption, and keep track of the backups by the commit hash so you can restore it when it rolls back.

rudehn:
@hongaar are you looking to set up an ORM such as [peewee](https://github.com/coleifer/peewee)? I think something like this would be beneficial since it abstracts out the DB instance from the data and provides a clean usage syntax.

andrewda:
GitHub user id would make it super easy to add profile pictures, too, so I'm all for that!

hongaar:
@md678685 yikes, yes I guess we need something like that. you think we also need offsite backups? s3? another gh repo?

@rudehn I have absolutely no experience with Python ORMs at all, so any help would be appreciated.

rudehn:
@hongaar Neither do I, but there's no better time to start learning. Using an ORM would abstract out all SQL and treat the db interactions in an object oriented fashion. It seems like an interesting concept, but not one we necessarily have to do.

flibustier:
It's going to be great !!
But I don't think an offsite backup is necessary, local as @md678685 said will be nice :)

hongaar:
@rudehn ORMs are great, can't live without [eloquent](https://laravel.com/docs/5.4/eloquent) in PHP, but don't feel confident enough to make the any architectural decisions on this in Python and don't have time atm to dive into docs... 😉

andrewda (approved):

PlasmaPower:
DB API merged, this is probably ready for continuation.

rudehn:
DB is in, this can be finished

PlasmaPower:
I'll start work on it.

PlasmaPower:
I've pushed my work to my branch `votes-persistence`. Feel free to pick it up. TODO:

- Migrations
- Keep track of user ID and username when counting votes (currently we just keep track of the username, but we need both for the DB)

rudehn:
@PlasmaPower, I don't mean to be pedantic, but what about this table layout?

```
 Table VotedOn
    user - User
    comment - Comment
    vote - int/bool
```

That way we can query what a user voted on, and who voted on a specific comment.

PlasmaPower:
@rudehn seems a bit excessive to me. We also couldn't easily migrate to that, we'd probably have to reset the votes.

rudehn:
Well I think the idea was to reset the votes anyways.

PR #393, opened 2017-05-30
Migrations

hongaar:
Migrations ([@rudehns idea](https://github.com/chaosbot/Chaos/pull/391#issuecomment-304885954))

- `startup.sh` runs `startup.d/50-run-migrations.sh`
- `startup.d/50-run-migrations.sh` runs `migrate.sh` and logs output
- `migrate.sh` iterates all files in `/migrations/` folder in alphabetical order and
  - checks whether the filename is already in the `.migrated` (could be moved to sqlite later) file and if not,
  - executes the file, and
  - saves the filename in the `.migrated` file

Furthermore:

- `callOnce` is removed (from #72)
- logs are stored in new `/log/` directory, moved supervisor logs here as well
- cleaned up `etc/nginx/chaos_errors`


etc/chaos_supervisor.conf
```diff
@@ -3,11 +3,11 @@ command=bash chaos_wrapper.sh
 directory=/root/workspace/Chaos
 user=root
 
-stdout_logfile=/var/log/supervisor/chaos-stdout.log
+stdout_logfile=/root/workspace/Chaos/log/supervisor-stdout.log
 stdout_logfile_maxbytes=1MB
 stdout_logfile_backups=10
 
-stderr_logfile=/var/log/supervisor/chaos-stderr.log
+stderr_logfile=/root/workspace/Chaos/log/supervisor-stderr.log
```
PlasmaPower, line 10: Does this require a change to the docker config?

hongaar: 👍

rudehn (approved):

PR #395, opened 2017-05-30
Contributor Demographic Diversity

Benyuel:
Based on the languages removed,added or byte per language changed in the pr diff, the contributor gets a multiplier towards their own vote in order to encourage diversity within the democracy.

### DIFF
--- a/github_api/prs.py
+++ b/github_api/prs.py
@@ -302,3 +302,12 @@ def post_status(api, urn, sha, state, description):
         "context": "chaosbot"
     }
     api("POST", path, json=data)
+
+
+def get_languages(api, urn, pr_num):
+    """ yield the difference in language diversity between this pr & master """
+    """ List languages for the specified repository. 
+        The value on the right of a language is the 
+        number of bytes of code written in that language. """
+    path = get_pr(api, urn, pr_num)['head']['repo']['url'] + '/languages'
+    return api("get", path)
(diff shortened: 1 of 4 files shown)


github_api/prs.py
```diff
@@ -312,3 +312,12 @@ def post_status(api, urn, sha, state, description):
...
     }
     api("POST", path, json=data)
+
+
+def get_languages(api, urn, pr_num):
+    """ yield the difference in language diversity between this pr & master """
+    """ List languages for the specified repository. 
+        The value on the right of a language is the 
+        number of bytes of code written in that language. """
+    path = get_pr(api, urn, pr_num)['head']['repo']['url'] + '/languages'
```
hongaar, line 322: why not re-use the `pr` object passed into `get_votes()`?

hongaar:
Like the implementation, but not sure if this is something which should count towards your vote weight, especially not with a 1:3 ratio...

mikesmiffy128:
This is definitely not something which should count. All this does is incentivise people to clutter the repository with random code in as many languages as possible for no reason. :-1:

Benyuel:
@mikesmiffy128: that's called chaos

mikesmiffy128:
Okay, but language diversity doesn't achieve anything. We're not even talking about culture here - that's a whole other debate - but this is literally just challenging people to shoehorn in code that fewer people will understand, probably making democracy more difficult due to the added effort of understanding multiple obscure languages.

There is literally no benefit to doing this.

Edit: sorry I have no sense of fun

PR #396, opened 2017-05-30
Allow meritocracy to change their reviews

PlasmaPower:
Currently, only negative -> positive is possible. Of course, if another commit was added, the old review would no longer count.

### DIFF
--- a/github_api/prs.py
+++ b/github_api/prs.py
@@ -97,8 +97,7 @@ def formatted_votes_short_summary(votes, total, threshold, meritocracy_satisfied
     meritocracy_str = "✓" if meritocracy_satisfied else "✗"
 
     return """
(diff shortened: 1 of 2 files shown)


PlasmaPower:
Please double check this, because I'm not 100% sure about some of the syntax used.

PlasmaPower:
I've also changed the status line to be a bit shorter. Previously it was:

> remaining: 3:10, vote: 4-0, weighted total: 4.0, threshold: 6.5, meritocracy: ✗

By default, GitHub shortens that to:

> remaining: 3:10, vote: 4-0, weighted total: 4.0, threshold: 6.5, meritocr...

Now it'll look like:

> remaining: 3:10, vote: 4-0 → 4.0, threshold: 6.5, meritocracy: ✗

syvb (approved):

phil-r (approved):

PR #389, opened 2017-05-30
Mission - find bugs

rudehn:
We are on a journey of finding out why the issue command system isn't working. Join me on this journey!

PR #399, opened 2017-05-30
Bug fix for fatal error (from #389)

rudehn:
Can't grab the length of a generator. This log line isn't needed anyways.


syvb (approved):

rudehn:
The hunt continues...

rudehn:
I found the error. As I suspected, there's no utilization of pagination in the API. We are requesting 100 comments per page, but only ever retrieve the first page of comments. The PR comments are also being considered "issue comments" and are spamming the returned results, leaving no commands recognized. Is this desired behavior?

I believe the solution is to fully implement pagination (in all our api calls that use it), and only retrieve issues  after a certain timestamp to prevent retrieving the entire comment history every 10 minutes.

PR #405, opened 2017-05-30
Add a "secret" terminal to the page

mdcfe:
Currently it's just a base for people to build off, but if someone implemented JSON-RPC or some more routes on the server side, we could have a neat way to access data from the server through this.


mdcfe:
Failing due to #389, not me.

rhengles:
I would approve this in a separate page, specially because it's not functional yet. It's adding dependencies on the home page, so for now it's 👎

o3LL:
3 libs for 1 funny moment 👎 
Do it from scratch 🗡

hongaar:
Frontend is mainly fun play and chaos anyway, so sure why not! 👍

mdcfe:
@rhengles The libraries are now loaded the first time the word "experiment" is clicked rather than immediately on page load, so it won't be loaded if you don't want to use it.

hongaar (approved):

o3LL (approved):

rhengles:
I would really rather have this on a separate page, but here's my upvote 👍  This can be improved later.

PR #417, opened 2017-05-31
Fix for closing stale issues

mark-i-m:
Github APIs use pagination, so we were getting the newest 30 issues, which are of course not usually stale...

### DIFF
--- a/cron/poll_issue_close_stale.py
+++ b/cron/poll_issue_close_stale.py
@@ -17,9 +17,11 @@ def poll_issue_close_stale(api):
     __log.info("Checking for stale issues...")
 
     # Get all issues
-    issues = gh.issues.get_open_issues(api, settings.URN)
+    issues = gh.issues.get_oldest_open_issues(api, settings.URN)
 
-    __log.info("There are currently %d open issues" % len(issues))
+    __log.info(str(issues))
+
+    __log.info("Got the oldest %d open issues" % len(issues))
 
     for issue in issues:
         number = issue["number"]
--- a/github_api/issues.py
+++ b/github_api/issues.py
@@ -16,9 +16,13 @@ def open_issue(api, urn, issue_id):
     return resp
 
 
-def get_open_issues(api, urn):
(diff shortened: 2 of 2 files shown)


PlasmaPower (approved):

PlasmaPower:
@mark-i-m in the future feel free to mention me or another member of the meritocracy if you need a review - I'll notice a lot sooner than a title change

andrewda:
@PlasmaPower how about having the bot request a review from the members of the meritocracy?

PlasmaPower:
@andrewda that could work. When would you propose that happen? When the voting window moves into the extended part? That'd mean a review would be an insta-merge.

andrewda:
Sure that would work. If there's not already a meritocracy approval, request reviews.

mark-i-m:
Alternately, chaosbot could randomly assign someone to every new PR

PlasmaPower:
@mark-i-m I don't think that's a good idea, as sometimes people are asleep or no longer active.

cron/poll_issue_close_stale.py
```diff
@@ -17,9 +17,11 @@ def poll_issue_close_stale(api):
     __log.info("Checking for stale issues...")
 
     # Get all issues
```
phil-r, line 19: this comment is no longer true ;)

phil-r (changes requested):

andrewda:
We're not going to be able to request reviews, because only people with write access can be requested (https://developer.github.com/v3/pulls/review_requests/). Maybe have chaosbot mention the list of meritocracy users? Not as clean, but our only choice I guess.

PlasmaPower:
@andrewda Yeah, I think that should happen whenever a PR enters the extended voting window.

mark-i-m:
This still doesn't seem to be working... I can't find the right spot in the log either...

mark-i-m:
Perhaps I am sorting in the wrong direction? I am kind of confused...

PR #419, opened 2017-05-31
Add links to voters list

evanmhm:
[🥒](https://www.compart.com/en/unicode/U+1F952)

### DIFF
--- a/server/static/js/chaos.js
+++ b/server/static/js/chaos.js
@@ -161,7 +161,7 @@ request.onreadystatechange = function() {
 
       var tablehtml = "<table>";
       for (var i = 0; i < list.length && i < 20; i++) {
-        tablehtml += "<tr><td><a href=\"https://github.com/" + list[i].names + "\">" + list[i].names + "</a></td><td>" + list[i].votes + "</tr>";
+        tablehtml += "<tr><td><a href=\"https://github.com/" + escape(list[i].names) + "\">" + list[i].names + "</a></td><td>" + list[i].votes + "</tr>";
       }
       tablehtml += "</table>";
       result.innerHTML = tablehtml;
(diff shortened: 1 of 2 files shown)


PlasmaPower:
To clarify, this links usernames in the interactive voter list.

andrewda (approved):

PlasmaPower:
Currently, this might have the potential for an XSS. Wrap the URL in quotes, then call `escape` on the username. @andrewda I suggest you revoke your review (post a neutral review after it) until this is fixed.

PlasmaPower:
Actually, there already was an XSS. I guess I'm :+1: on this change, but please also fix the XSS.

PlasmaPower:
I'm not sure that's possible with a GitHub username, but better safe than sorry. Maybe some time of weird space messes things up too.

server/static/js/chaos.js
```diff
@@ -161,7 +161,7 @@ request.onreadystatechange = function() {
 
       var tablehtml = "<table>";
       for (var i = 0; i < list.length && i < 20; i++) {
-        tablehtml += "<tr><td>" + list[i].names + "</td><td>" + list[i].votes + "</tr>";
+        tablehtml += "<tr><td><a href=\"https://github.com/" + list[i].names + "\">" + list[i].names + "</a></td><td>" + list[i].votes + "</tr>";
```
PlasmaPower: Could you also replace `list[i].names` with `escape(list[i].names)`?

PlasmaPower: Oops that's for URIs. Don't do that.

PlasmaPower: I'd go with this: https://stackoverflow.com/a/12034334/3352674

PlasmaPower: NVM that's correct for the URL.

PlasmaPower (approved):
Looks good to me now.

PlasmaPower:
Looks like GH usernames are alphanumeric, so there's no reason for worry. Sorry for the confusion!

PR #422, opened 2017-05-31
Ability to run tests in docker too

mark-i-m:
The docker env already has all of the deps and stuff...

### DIFF
--- a/dev/docker/test.bat
+++ b/dev/docker/test.bat
@@ -1,2 +1,2 @@
 cd ..\..\
-docker run -it --rm -v %cd%:/root/workspace/Chaos -p 8082:80 -p 8081:8081 chaos python -m unittest
+docker run -it --rm -v %cd%:/root/workspace/Chaos -p 8082:80 -p 8081:8081 chaos python -m unittest discover -p "*_test.py"
--- a/dev/docker/test.bat
+++ b/dev/docker/test.bat
@@ -1,2 +1,2 @@
 cd ..\..\
-docker run -it --rm -v %cd%:/root/workspace/Chaos -p 8082:80 -p 8081:8081 chaos python -m unittest discover -p "*_test.py"
+docker run -it --rm -v %cd%:/root/workspace/Chaos -p 8082:80 -p 8081:8081 chaos bash /root/workspace/Chaos/dev/docker/start_tests.sh
--- a/dev/docker/test.bat
+++ b/dev/docker/test.bat
@@ -1,0 +1,2 @@
+cd ..\..\
+docker run -it --rm -v %cd%:/root/workspace/Chaos -p 8082:80 -p 8081:8081 chaos python -m unittest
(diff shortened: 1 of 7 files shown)


PlasmaPower (approved):

flibustier (approved):

dev/docker/test.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+docker run -it --rm\
+    -v $(cd ../../; pwd):/root/workspace/Chaos\
+    -p 8082:80\
+    -p 8081:8081\
+    chaos\
+    python -m unittest
```
phil-r: This line should be `python -m unittest discover -p "*_test.py"`
Edit: added missing quote

phil-r (changes requested):

dev/docker/test.bat
```diff
@@ -0,0 +1,2 @@
+cd ..\..\
+docker run -it --rm -v %cd%:/root/workspace/Chaos -p 8082:80 -p 8081:8081 chaos python -m unittest
```
phil-r: here `discover -p "*_test.py"` is also missing
Edit: added missing quote

phil-r (changes requested):

dev/docker/test.sh
flibustier: @phil-r is right, I just tested! (Otherwise it makes a mistake)
And it's `python -m unittest discover -p *_test.py` without the `"` in @phil-r's answer :)

dev/docker/test.bat
flibustier: Remark applicable here as well :p

dev/docker/test.sh
flibustier: @phil-r is right, I just tested! (Otherwise it makes a mistake)
And it's `python -m unittest discover -p *_test.py` without the `"` in @phil-r's answer :)

flibustier: @phil-r is right, I just tested! (Otherwise it makes a mistake)
And it's `python -m unittest discover -p *_test.py` without the `"` in @phil-r's answer :)

PlasmaPower: No @phil-r is right it does need quotes. The argument is intended for `unittest`, without the quotes bash tries to expand it and fails.

mark-i-m: Thanks, I was pretty sure it was wrong

mark-i-m: What I have, that is

mark-i-m:
I think I can un-WIP if it looks good to everyone

PlasmaPower (approved):
Looks good to me.

mark-i-m:
Hmm... I am going to add `flake8` to the test run too...

PlasmaPower:
@mark-i-m sure, why not.

PlasmaPower (approved):

PR #428, opened 2017-05-31
Update README in anticipation of #422

flibustier:

PR #430, opened 2017-05-31
33% more efficient and a missile launcher inside 🐲

flibustier:
This PR has nothing to do with a missile launcher or a dragon, sorry.

BUT now we have a docker-compose file to simply run Chaos and tests (so it's based on #422 an #428)
- docker-compose is simple for running docker commands
- crossplatform : no need of 6 scripts (run.sh, tests.bat, etc…)

Only 2 commands to rule them all :
- `docker-compose up tests` will run… tests
- `docker-compose up chaos` will run… Chaos!

### DIFF
--- a/dev/docker/docker-compose.yml
+++ b/dev/docker/docker-compose.yml
@@ -3,7 +3,7 @@ services:
   chaos:
     image: chaos
     build:
-      context: .
+      context: ../..
       dockerfile: ./dev/docker/Dockerfile
     command: bash /root/workspace/Chaos/dev/docker/start_services.sh
     links:
@@ -12,16 +12,16 @@ services:
       - 8082:80
       - 8081:8081
     volumes:
-      - .:/root/workspace/Chaos
-
+      - ../..:/root/workspace/Chaos
+      
   tests:
     image: chaos
     build:
-      context: .
+      context: ../..
       dockerfile: ./dev/docker/Dockerfile
     command: python -m unittest discover -p *_test.py
     volumes:
-      - .:/root/workspace/Chaos
+      - ../..:/root/workspace/Chaos
 
   db:
      image: mysql:5.7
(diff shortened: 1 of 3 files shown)


Moussto (approved):

andrewda (approved):

numaru:
👍

We could move the `docker-compose.yml` file to `dev/docker` and use `context: ../../`.
Also, both images are named `chaos`.

flibustier:
> Also, both images are named chaos.

Yes, because it's the same image so docker compose will name the image "chaos" (avoiding 2 builds)

PlasmaPower (approved):

dev/docker/docker-compose.yml
```diff
@@ -0,0 +1,38 @@
+version: '3'
+services:
+  chaos:
+    image: chaos
+    build:
+      context: ../..
+      dockerfile: ./dev/docker/Dockerfile
+    command: bash /root/workspace/Chaos/dev/docker/start_services.sh
+    links:
```
DennyLoko, line 9: Why `links`? You should be using the Docker network and the `depends_on` key.

flibustier: Yes you're right! An old habit… I will fix it!
(`db` is not used for now but it will serve in future)

dev/docker/docker-compose.yml
```diff
@@ -0,0 +1,38 @@
+version: '3'
+services:
+  chaos:
+    image: chaos
+    build:
+      context: ../..
+      dockerfile: ./dev/docker/Dockerfile
+    command: bash /root/workspace/Chaos/dev/docker/start_services.sh
```
phil-r, line 8: @flibustier What's `start_services.sh` ?

numaru: It was an anticipation of #422 merging.

phil-r: I see, thanks @numaru ;)

PR #427, opened 2017-05-31
Prevent animation on first card click

mdcfe:
Currently it animates to show the main card when you click on it after loading the page, even though it's already showing the main card.

PR #425, opened 2017-05-31
Fix center and responsive website.

Alexays:
Fix firefox center.
On mobile cards take 85% of width to allow to click on other.

PR #431, opened 2017-05-31
Fix center and responsive again.

Alexays:
Same as #425.
Fault of Meritocracy :o
edit: Do not Animate on load, #427 ok


andrewda:
I'll approve when Travis passes 👍

Alexays:
They seem to have no travis test this time.

andrewda:
It was hanging, but the test it just started.

andrewda:
Have no idea why it didn't generate a status but it passed: https://travis-ci.org/chaosbot/Chaos/builds/237924313

andrewda (approved):

PlasmaPower (approved):

andrewda (approved):

server/static/js/chaos.js
```diff
@@ -172,35 +174,35 @@ request.onreadystatechange = function() {
...
-    } else if (showTerminal) {
-        showTerminal();
-    }
-}
+  if (!termLoading) {
+    loadJq();
+  } else if (showTerminal) {
+    showTerminal();
+  }
+}
```
PlasmaPower: Please add a newline here.

server/static/js/chaos.js
```diff
@@ -172,35 +174,35 @@ request.onreadystatechange = function() {
...
-        loadJq();
-    } else if (showTerminal) {
-        showTerminal();
-    }
-}
+  if (!termLoading) {
+    loadJq();
+  } else if (showTerminal) {
+    showTerminal();
+  }
```
andrewda: I don't feel great about changing all the spacing in this PR. Seems like something that would fit better in another PR.

andrewda (changes requested):

server/static/js/chaos.js
mdcfe: I was considering adding ESLint (most likely with Airbnb rules), although that might be more trouble than it's worth.

o3LL (approved):

server/static/js/chaos.js
andrewda: If it means we could lint it, I'm all for it!

phil-r:
This happens to travis, just ignore it..

server/static/js/chaos.js
Alexays: Sorry for this, reflex x)

rhengles: If we must enforce a style guide, I suggest https://github.com/prettier/prettier !

andrewda (approved):
good now 👍

server/static/js/chaos.js
mdcfe: Prettier doesn't seem to lint though, it formats the code for you. It won't do anything to code which breaks the style guide before it is merged into the repo, unless we run it while merging.

andrewda (approved):

o3LL (approved):

mdcfe (approved):

PR #434, opened 2017-05-31
Retry on 202 status code

### LINKED ISSUE #432: Meritocracy list bug ?
-> https://github.com/chaosbot/Chaos/pull/425#issuecomment-305189271

andrewda:
Fixes #432 

_(probably)_

### DIFF
--- a/github_api/__init__.py
+++ b/github_api/__init__.py
@@ -95,7 +95,7 @@ class API(object):
         # wait 5 seconds, then retry request due to 202 status code
         if resp.status_code == 202:
             time.sleep(5)
-            self.__call__(method, path, **kwargs)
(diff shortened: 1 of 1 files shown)


rudehn:
What about the response data? Looks like you're just dropping it

andrewda:
Just tested and it's working as advertised. Un-WIPing.

PlasmaPower:
@rudehn On a 202 the response data should be empty or not important.

rudehn:
Right, but after we wait and requery, the result was never returned. The requery is not guaranteed to be a 202.

andrewda:
Yea that was a typo - should be working properly now.

PlasmaPower (approved):

PR #436, opened 2017-05-31
Give me all your lovin' ..err votes

rudehn:
The existing file is corrupted. Remove the lingering data so the service can run.

I'm not sure if the migration tool is working yet, so just putting the patch here.

The code is working, you can see detected commands [here](http://chaosthebot.com/issue_commands_ran.json), however one item is corrupted and has to be removed.

Last PR for this to work I swear.


PlasmaPower (approved):

mdcfe (approved):

hongaar:
@rudehn migration tool should be up and running, but never used yet... Just put a new file (with shebang) on `migrations` folder.

rudehn:
😄 it works!

PR #439, opened 2017-05-31
Implement temporary error page

### LINKED ISSUE #365: Thematic 404 page 🎭
There's a lack of pages on the site, but does that mean we deserve a default 404 page?

mdcfe:
Should suffice until we get an actual request handler (if we even get a proper request handler). Closes #365.

### DIFF
--- a/server/error.html
+++ b/server/error.html
@@ -1,0 +1,31 @@
+<!DOCTYPE html>
+<html lang="en">
+    <head>
+        <title>Error – ChaosBot</title>
+        <meta charset="utf-8">
+    </head>
+    <body>
+        <p>The page could not be loaded.</p>
+        <p><a href="/">Return to main page</a></p>
+
+        <style>
+            body {
+                height: 100vh;
+                display: flex;
+                font-family: sans-serif;
+                text-align: center;
+                flex-direction: column;
+                align-items: center;
+                justify-content: center;
+            }
+
+            body > p:first-child {
+                font-size: x-large;
+            }
+
+            body > p {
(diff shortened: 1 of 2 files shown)


PlasmaPower:
Will this also apply for `404`s? Is any other error possible?

mdcfe:
@PlasmaPower This applies to all errors, including 404s. I haven't used any Python web frameworks for quite a while, so I didn't want to just go and add one without everyone's approval.

~~If it's possible to override the `SimpleHTTPRequestListener.send_error` function then that would work better, but I'm not quite sure how that would work.~~ See below.

mdcfe:
Correction: after looking through the `send_error` function, we can use `code`, `message` and `explain`. I'll add those to the page now.

o3LL (approved):

mdcfe:
@PlasmaPower This should now substitute in any errors with their status code, message and explanation.

Example:
chrome_2017-05-31_22-35-40

PlasmaPower (approved):

RyanZim (approved):

hongaar:
@md678685 

image

Bug or feature?

mdcfe:
@hongaar Just went to see what happens in production before I saw your comment - it worked fine locally with Python 3.6.1, not quite sure what happened there 😆

mdcfe:
Turns out using Vagrant installs Python 3.4. In 3.5 `%(code)s` changed, and the production environment runs 3.6. Going to make a PR to fix Vagrant, and a separate one to fix the error.

Issue #364, opened 2017-05-29
Chaosbot can makes tweets


eamanu:
What do you think about the idea to make that Chaosbot tweets when:

* There is a new PR
* When merge
* When somebody open a new Issue

phil-r:
- when it goes down 💥

chaosbot:
This issue hasn't been active for a while.To keep it open, react with :-1: on the `vote close` post.

chaosbot:
> /vote close

Vote Failed

chaosbot:
> /vote close

Vote Failed

chaosbot:
> /vote close

Vote Failed

chaosbot:
> /vote close

Vote Failed

chaosbot:
> /vote close

Vote Failed

chaosbot:
> /vote close

Vote Failed

chaosbot:
> /vote close

Vote Failed

chaosbot:
> /vote close

Vote Failed

chaosbot:
> /vote close

Vote Failed

chaosbot:
> /vote close

Vote Failed

chaosbot:
> /vote close

Vote Failed

chaosbot:
> /vote close

Vote Failed

chaosbot:
> /vote close

Vote Failed

chaosbot:
> /vote close

Vote Failed

chaosbot:
> /vote close

Vote Failed

chaosbot:
> /vote close

Vote Failed

chaosbot:
> /vote close

Vote Failed

chaosbot:
> /vote close

Vote Failed

chaosbot:
> /vote close

Vote Failed

chaosbot:
> /vote close

Vote Failed

chaosbot:
> /vote close

Vote Failed

chaosbot:
> /vote close

Vote Failed

chaosbot:
> /vote close

Vote Failed

chaosbot:
> /vote close

Vote Failed

chaosbot:
> /vote close

Vote Failed

chaosbot:
> /vote close

Vote Failed

chaosbot:
> /vote close

Vote Failed

chaosbot:
> /vote close

Vote Failed

[...]

PR #443, opened 2017-06-01
Add a Twitter Api according to #364 - need testing

eamanu:
The `twitter_api.Twitter.PostTwitter(message)` receive a message and make a Tweet
Is necessary have a `twitter_api.secre`t like `github_pat.secret` with the next format
```
consumer_key='1234...'
consumer_secret='1234...'
access_token='12345...'
access_secret='1234...'
```

Todo List:
1. Test twitter_api
1. create a Twitter Apps (https://apps.twitter.com/) for ChaosBot and complete the twitter_api.secret


rudehn:
I would recommend putting your twitter keys into a json file. Then loading them is as simple as `keys = json.load(file_pointer)`

phil-r:
To add token/secret we can use `encryption` module added in #227
Documentation is here: https://github.com/chaosbot/Chaos/pull/321/files (never was merged 😢)

hongaar:
@phil-r would be nice if we could add this to the wiki!

phil-r:
@hongaar added: https://github.com/chaosbot/Chaos/wiki/Encryption

syvb (approved):

bobjoe27326:
:+1:

eamanu:
@phil-r I don't undestand how can I use the encryptation with the Twitter's keys. Can you make a little sample?

syvb (approved):

Swizz:
Is Chaosbot :-1: referenced as a negative vote ? Oo

chaos.py
```diff
@@ -69,12 +74,17 @@ def main():
 
     api = gh.API(settings.GITHUB_USER, settings.GITHUB_SECRET)
 
+    # Api Twitter
+    api_twitter = ta.API_TWITTER(settings.TWITTER_API_KEYS_FILE)
+
     log.info("checking if I crashed before...")
+    ta.Twitter.PostTwitter("checking if I crashed before...", api_twitter.GetApi())
```
andrewda, line 81: Do we really want it to be this verbose? I don't think it needs to tweet this out every time it starts up.

chaos.py
```diff
@@ -69,12 +74,17 @@ def main():
...
 
     # check if chaosbot is not on the tip of the master branch
     check_for_prev_crash(api, log)
 
     log.info("starting up and entering event loop")
+    ta.Twitter.PostTwitter("starting up and entering event loop", api_twitter.GetApi())
```
andrewda, line 87: Same as above.

andrewda (changes requested):
Maybe a tweet when a PR gets merged, or created. Maybe also change the current tweets to "Starting up..." instead of two simultaneous tweets when starting up.

rudehn:
I think anything posted to twitter should be important. The info posted right now is better suited to log files

eamanu:
@rudehn @andrewda I will make that Chaosbot make tweets when there are  PR or Merges. What are the files that make the PR or files of interest to change?

chaos.py
eamanu: I think like @rudehn. Anything posted to twitter should be important. A post in twitter isn't a cost or some extra.

chaos.py
eamanu: Same as above :+1:

PlasmaPower:
@eamanu It probably should tweet when any PR gets merged. Alternatively, we could ignore PRs with only documentation changes (only `*.md` files touched).

eamanu (commented):
I think like @rudehn. Anything posted to twitter should be important. A post in twitter isn't a cost or some extra

mark-i-m:
I suspect that posting crashes might be good...

dbpokorny:
> I suspect that posting crashes might be good...

https://www.youtube.com/watch?v=bQlpDiXPZHQ

Life Call Commercial "I've fallen and I can't get up!"

> I'm having interpreter pain!

> I've halted, and I can't restart!

PR #444, opened 2017-06-01
Solve I101 and I201 format bug

eamanu:
I101: The names in your from import are in the wrong order.
I201: Missing newline between sections or imports.

All this according with 'flake8-import-order'


PlasmaPower:
This is from a flake8 extension, right? Mind adding that extension to Travis and Docker?

phil-r:
This change should include whatever requested this change.. otherwise it's really strange

PR #446, opened 2017-06-01
Fix docker build by installing six before req's

drewrey:
Previously, `docker-compose build` in `/dev/docker/` was failing due to
`ModuleNotFoundError`s for six.

### DIFF
--- a/dev/docker/Dockerfile
+++ b/dev/docker/Dockerfile
@@ -47,6 +47,7 @@ ENV PATH="$venv/bin:$PATH"
 
 WORKDIR $chaosdir
 
+RUN pip install six
 COPY requirements.txt .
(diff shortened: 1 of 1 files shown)


drewrey:
Ah, this PR should also probably update the Vagrantfile?

But then, it would be nice to have one place to specify this order of installation. It seemed to me as though six needed to be installed separately and prior to `pip install -r requirements.txt`, but perhaps there's some way around that?

numaru:
I'm running docker over b2d and I failed to reproduce 😿.
You could try with `docker-compose build --no-cache tests`.

phil-r:
It looks like this fixed docker build for me

phil-r (approved):

PR #454, opened 2017-06-01
Update/add missing dependencies to Vagrantfile

mdcfe:
This should reduce the chance of version differences causing errors (such as in #439).

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "python.pythonPath": "C:/Program Files (x86)/Python36-32/python.exe"
+}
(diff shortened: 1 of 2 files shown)


syvb (approved):

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "python.pythonPath": "C:/Program Files (x86)/Python36-32/python.exe"
```
flibustier: Hmmm, this is need to be commited?

syvb (changes requested):
Remove .vscode

.vscode/settings.json
mdcfe: Oops, that shouldn't have been committed. I thought it had been removed beforehand.

PlasmaPower (approved):

PR #455, opened 2017-06-01
FIX ERROR PAGES WITH THIS ONE SIMPLE TRICK

mdcfe:
This fixes the display of the error code in #439 - the previous page worked on Python 3.4 (installed by default by Vagrant until #454), but the way error codes work for SimpleHTTPRequestListener changed in Python 3.5.

### DIFF
--- a/server/error.html
+++ b/server/error.html
@@ -5,7 +5,7 @@
         <meta charset="utf-8">
     </head>
     <body>
-        <p id="code">%(code)s<p>
+        <p id="code">%(code)d<p>
         <p id="message">%(message)s</p>
         <p id="explain">%(explain)s</p>
         <p id="link"><a href="/">Return to main page</a></p>


syvb (approved):

o3LL:
Damn, we are not on youtube 😆

o3LL (approved):

phil-r (approved):

mdcfe:
Doesn't seem like ChaosBot is counting the right number of votes - shouldn't it be at least 9 positive not 5 (mine + 8 👍s)?

PlasmaPower:
@md678685 Some votes are probably from accounts < 1 month old.

rudehn:
@PlasmaPower what if we change that rule? Any account < 1 month old that submits a PR and is accepted would have their vote count

PlasmaPower:
@rudehn I'll submit a PR for that.

mdcfe:
It still hasn't changed - is it broken?

andrewda:
Yea - one of the earlier PR's broke it. It keeps crashing due to the labels API returning a `422` status and the bot not accounting for that properly http://chaosthebot.com:8081/chaos-stderr.log

rudehn:
Looks like it's time for another hotfix. Here's an idea, we automatically email @amoffat when chaosbot had a fatal error 😆

andrewda:
Or text him ;)

amoffat:
lol.  yeah there is no real good way for me to know.  i have the logs constantly scrolling and check it every few hours

PR #456, opened 2017-06-01
Database abstraction layer (sqlite implementation)

hongaar:
```python
db = DB.get_instance()  # type: DB

db.query("DROP TABLE IF EXISTS users ")
db.query("CREATE TABLE users (id INTEGER PRIMARY KEY AUTOINCREMENT NOT NULL, name TEXT NOT NULL)")
db.query("INSERT INTO users (name) VALUES (?)", ['test1'])
db.query("INSERT INTO users (name) VALUES (?)", ['test2'])
db.query("INSERT INTO users (name) VALUES (?)", ['test3'])

print(db.query("SELECT * FROM users"))
print(db.row("SELECT * FROM users"))
print(db.one("SELECT * FROM users"))
```
⏬ 
```
[(1, 'test1'), (2, 'test2'), (3, 'test3')]
(1, 'test1')
1
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,4 @@ api_cache/
 server/issue_commands_ran.json
 .migrated
 ansible/*.retry
+db.sqlite
--- a/lib/db.py
+++ b/lib/db.py
@@ -1,0 +1,45 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+from importlib import import_module
+
+import settings
+from lib.db_adapters.interface import AbstractAdapter
+
+
+class DB(object):
+    """ our database abstraction layer """
+
+    instance = None  # type: DB
+
+    @staticmethod
+    def get_instance():
+        """ singleton implementation """
+        if DB.instance is None:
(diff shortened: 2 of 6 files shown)


flibustier:
The PR we all expected! Great :)

flibustier:
Well done!!
Next step : Blockchain :D

mdcfe:
The web server currently runs in its own process, and as far as I know so do the cron jobs? I think we should go with MySQL, purely because it's practically guaranteed that we can modify it concurrently without interfering with other processes or having to wait for locks.

hongaar:
@md678685 you're in luck: https://github.com/chaosbot/Chaos/pull/461

_edit_ if you vote that one up, you'll be implicitly also voting this one up 😄

rudehn:
I'm going to add a hook to the DB class to support an ORM as well. I'd just prefer not having to rewrite any queries and have the ORM do it all for me  😄

PlasmaPower:
We're just merging the mysql PR as it includes this one, right? Or was this one supposed to go through too?

PR #461, opened 2017-06-01
Database abstraction layer (mysql implementation)

hongaar:
Includes #456 
Based on #381

### DIFF
--- a/settings.py
+++ b/settings.py
@@ -101,3 +101,10 @@ DB_ADAPTER = "sqlite"
 DB_CONFIG = {
     "filename": "db.sqlite"
 }
+# MySQL example
+# DB_CONFIG = {
+#     "host ": "localhost",
+#     "user ": "chaos",
+#     "password ": "chaos",
+#     "db ": "db"
+# }
--- a/settings.py
+++ b/settings.py
@@ -95,3 +95,9 @@ MERITOCRACY_TOP_CONTRIBUTORS = 10
 
 # The top n voters will be allowed in the meritocracy
 MERITOCRACY_TOP_VOTERS = 10
+
+# Database settings
+DB_ADAPTER = "sqlite"
+DB_CONFIG = {
+    "filename": "db.sqlite"
+}
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,4 @@ api_cache/
 server/issue_commands_ran.json
 .migrated
 ansible/*.retry
+db.sqlite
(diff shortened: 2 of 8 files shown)


rudehn:
The mysql implementation will return a dict cursor instead of a tuple cursor. Might want to standardize that

hongaar:
@rudehn 👍 see  https://github.com/chaosbot/Chaos/pull/456/commits/a1038b94686b04199f1b62eed193e44c292350be

settings.py
```diff
@@ -95,3 +95,16 @@
...
+DB_CONFIG = {
+    "filename": "db.sqlite"
+}
+# MySQL example
+# DB_CONFIG = {
+#     "host ": "localhost",
+#     "user ": "chaos",
+#     "password ": "chaos",
+#     "db ": "db"
+# }
```
PlasmaPower, line 110: Shouldn't we be using MySQL? Transitioning will be difficult.

hongaar:
@PlasmaPower I'd prefer sqlite for now, because it's much faster and more portable for simply storing some simple state records.

@md678685 With MySQL you'll also have to wait for locks; the operations are just running in another process.

And with MySQL we'll get another server to monitor and another dependency to manage (plus a whole bunch of memory goes to waste).

syvb (approved):

PlasmaPower:
@hongaar How would you recommend I approach creating the table? I could do it with the sqlite CLI and the migrations tool.

rudehn:
@PlasmaPower Just a heads up, but I'm going to start working on integrating [peewee](https://github.com/coleifer/peewee) so we can abstract out the db type from our queries. Peewee supports both mysql and sqlite.

If you'd like to take a look into it, you're more than welcome. I think having an ORM would be useful.

PlasmaPower:
@rudehn Okay, I might wait for that. Will you be introducing a migration strategy? It looks like that has functions for doing the migrations themselves, but not versioning or similar.

rudehn:
I envisioned using the migrations tool and peewee's migration functions. I don't see any versioning either. Is that something we are wanting to do?

PlasmaPower:
@rudehn I guess the existing tools will work, it's just that they're shell based. I forgot that we could run a separate Python script from them.

rudehn:
You can go ahead and setup a python migration script to get started.

PlasmaPower:
@rudehn there's a problem - you can't import `lib.db` from the a script in the migrations directory.

rudehn:
maybe we need to add the root level to the path? Try

`
sys.path.insert(...)
`

PlasmaPower:
That might work, but it's pretty hacky.

PR #462, opened 2017-06-01
Dump meritocracy data alongside voters.json

mdcfe:
Will be dropped in the root of the web server when generated.

### DIFF
--- a/cron/poll_pull_requests.py
+++ b/cron/poll_pull_requests.py
@@ -51,7 +51,7 @@ def poll_pull_requests(api):
         meritocracy = top_voters | top_contributors
         __log.info("generated meritocracy: " + str(meritocracy))
 
-        with open('server/meritocracy.json', 'r+') as mfp:
+        with open('server/meritocracy.json', 'w') as mfp:
             json.dump(meritocracy, mfp)
 
         needs_update = False
--- a/cron/poll_pull_requests.py
+++ b/cron/poll_pull_requests.py
@@ -30,6 +30,12 @@ def poll_pull_requests(api):
         # file already exists, which is what we want
         pass
 
+    try:
+        mfp = open('server/meritocracy.json', 'x')
+        mfp.close()
+    except:
+        pass
+
     with open('server/voters.json', 'r+') as fp:
         total_votes = {}
         fs = fp.read()
@@ -44,6 +50,9 @@ def poll_pull_requests(api):
         top_voters = set([user.lower() for user in top_voters[:settings.MERITOCRACY_TOP_VOTERS]])
         meritocracy = top_voters | top_contributors
         __log.info("generated meritocracy: " + str(meritocracy))
+
+        with open('server/meritocracy.json', 'r+') as mfp:
+            json.dump(meritocracy, mfp)
 
         needs_update = False
         for pr in prs:
--- a/cron/poll_pull_requests.py
+++ b/cron/poll_pull_requests.py
@@ -30,12 +30,6 @@ def poll_pull_requests(api):
         # file already exists, which is what we want
         pass
 
-    try:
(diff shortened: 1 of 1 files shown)


davidak (approved):

PlasmaPower:
This will append the new meritocracy json to the old one, similar to the broken voters json issue. Why not just open it once in `w` mode?

@davidak rescind review?

rudehn:
>  ``r+''  Open for reading and writing.  The stream is positioned at the
         beginning of the file.

That doesn't sound good... Why not just use `"r"` and `"w"`?

PlasmaPower:
Why not just `w`? We never read from it.

I think `r+` would work if we called truncate before writing to it.

PlasmaPower:
And yes, this won't append to the file, but it will cause problems. Example:

Write 1: abc
Write 2: de

Result: dec

That's because we don't truncate the file, so old data gets left behind.

rudehn:
@PlasmaPower I mean use `"r"` and `"w"` in a general sense.

mdcfe:
Oops, I'll amend that now. Forgot about file modes 😟

cron/poll_pull_requests.py
```diff
@@ -30,6 +30,12 @@ def poll_pull_requests(api):
         # file already exists, which is what we want
         pass
 
+    try:
+        mfp = open('server/meritocracy.json', 'x')
```
PlasmaPower: I don't see the point in this. The later call to `open` with `w` will already create the file if it doesn't exist.

PlasmaPower (changes requested):

andrewda (approved):
Looks good now!

phil-r:
Good job:
```
chaos_1  | Traceback (most recent call last):
chaos_1  |   File "chaos.py", line 120, in <module>
chaos_1  |     main()
chaos_1  |   File "chaos.py", line 70, in main
chaos_1  |     schedule.run_pending()
chaos_1  |   File "/root/.virtualenvs/chaos/lib/python3.6/site-packages/schedule/__init__.py", line 462, in run_pending
chaos_1  |     default_scheduler.run_pending()
chaos_1  |   File "/root/.virtualenvs/chaos/lib/python3.6/site-packages/schedule/__init__.py", line 75, in run_pending
chaos_1  |     self._run_job(job)
chaos_1  |   File "/root/.virtualenvs/chaos/lib/python3.6/site-packages/schedule/__init__.py", line 129, in _run_job
chaos_1  |     ret = job.run()
chaos_1  |   File "/root/.virtualenvs/chaos/lib/python3.6/site-packages/schedule/__init__.py", line 377, in run
chaos_1  |     ret = self.job_func()
chaos_1  |   File "/root/workspace/Chaos/cron/__init__.py", line 11, in <lambda>
chaos_1  |     lambda: poll_pull_requests(api))
chaos_1  |   File "/root/workspace/Chaos/cron/poll_pull_requests.py", line 49, in poll_pull_requests
chaos_1  |     json.dump(meritocracy, mfp)
chaos_1  |   File "/usr/lib/python3.6/json/__init__.py", line 179, in dump
chaos_1  |     for chunk in iterable:
chaos_1  |   File "/usr/lib/python3.6/json/encoder.py", line 437, in _iterencode
chaos_1  |     o = _default(o)
chaos_1  |   File "/usr/lib/python3.6/json/encoder.py", line 180, in default
chaos_1  |     o.__class__.__name__)
chaos_1  | TypeError: Object of type 'set' is not JSON serializable
```

phil-r:
This crashes chaos for me with `TypeError: Object of type 'set' is not JSON serializable`

PlasmaPower:
Looks like we need to turn it into a list first. Fixing.

PR #465, opened 2017-06-01
Lottery system

OffensiveCheese:
Some people will have more power than others, and that's OK. Depends how lucky you are.

### DIFF
--- a/github_api/voting.py
+++ b/github_api/voting.py
@@ -140,7 +140,7 @@ def get_vote_sum(api, votes):
         weight = get_vote_weight(api, user)
         if random.randint(0, 500) == 42:
             # Jackpot!
-            weight = weight * 100
+            weight = weight * 2.0
         weighted_vote = weight * vote
         total += weighted_vote
         if weighted_vote > 0:
--- a/github_api/voting.py
+++ b/github_api/voting.py
@@ -1,4 +1,5 @@
 import arrow
+import random
 from emoji import demojize
 
 from github_api.misc import dynamic_voting_window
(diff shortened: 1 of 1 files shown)


mikesmiffy128:
I'm no python expert, but I'd guess that the RNG is probably exploitable if it's not meant for crypto stuff. Also, this is wrong on principle anyway.

OffensiveCheese:
It's automatically seeded with system time. should be fine

mikesmiffy128:
Seeding with system time is like the worst you can do! :/

OffensiveCheese:
That's apart of the fun

PlasmaPower:
:-1: whenever people make RNG PRs, they forget to look at the mistakes of the past. This means pretty much every PR will be merged.

github_api/voting.py
```diff
@@ -137,6 +138,9 @@ def get_vote_sum(api, votes):
     negative = 0
     for user, vote in votes.items():
         weight = get_vote_weight(api, user)
+        if random.randint(0,500) == 42:
+            #Jackpot!
+            weight = weight * 100
```
mdcfe: `weight * 100` is a drastic increase - if this is going to be implemented, it should be much lower, like `* 1.1`, or `* 1.5` at the most.

mdcfe: Also, the bot should include in the status check whether anyone hit the jackpot.

mdcfe: Correct me if I'm wrong, but isn't this recalculated every time the votes are polled anyway?

mdcfe (changes requested):

github_api/voting.py
rudehn: yup. This idea has been proposed a lot in the past

nstanard (changes requested):
👿

PR #466, opened 2017-06-01
Change meritocracy distribution

PlasmaPower:
I think this will be a bit better, as IMO contributing is more important than voting

### DIFF
--- a/settings.py
+++ b/settings.py
@@ -100,10 +100,10 @@ CHAOSBOT_STDERR_LOG = join(THIS_DIR, "log/supervisor-stderr.log")
 ISSUE_STALE_THRESHOLD = 60 * 60 * 24 * 3  # 3 days
 
 # The top n contributors will be allowed in the meritocracy
-MERITOCRACY_TOP_CONTRIBUTORS = 10
+MERITOCRACY_TOP_CONTRIBUTORS = 12
 
 # The top n voters will be allowed in the meritocracy
-MERITOCRACY_TOP_VOTERS = 10
+MERITOCRACY_TOP_VOTERS = 5
 
 # Database settings
 DB_ADAPTER = "sqlite"


andrewda:
I think we should keep the total at 20 (because in reality it would be more like 10-12 after the duplicates). 12 and 8 would probably be good, or 14 and 6.

PlasmaPower:
@andrewda Overlaps reduce that though - I'll calculate what change this'll have.

PlasmaPower:
This would reduce the size of the current meritocracy, but only from 15 to 14.

andrewda:
Ok that works for me. But I think that mentioning the members of the meritocracy when entering the extended voting period is a must, especially because his will change the meritocrats.

PlasmaPower:
@andrewda I'd prefer to do that after the DB PR gets merged. It'll be a lot more stable then.

andrewda:
Yep works for me (as long as it's done pretty soon)

PlasmaPower:
Merge conflicts fixed. I'll start working on the meritocracy mentions now.

rudehn:
I think the meritocracy balance might need to be tweaked. I'm not sure the top 5 voters is useful, since pretty much all of them are already in the meritocracy

andrewda (approved):

PlasmaPower:
@rudehn Pretty much all, but not all. That could also change. IMHO top voters isn't useful, but I added it due to popular approval.

hongaar:
I'd say voters should be in there somehow, maybe even number of comments on issues/PRs? That way we 'reward' contributions from the community who doesn't actually write code.

andrewda:
That's a good point @hongaar - something like the judicial branch of the Chaosbot democracy

PlasmaPower:
@hongaar I'd support that, but the implementation might be complicated.

andrewda:
@PlasmaPower it would be much easier when database stuff gets up and running

PlasmaPower:
One vote needed.

PlasmaPower:
Or maybe 2, there's a decimal behind the `.5`.

PlasmaPower:
@anythingbot well it's a good think I don't care :)

PlasmaPower:
@anythingbot that'd be a republic, not a meritocracy. I've also considered it, but I think the current solution works fine.

PR #475, opened 2017-06-02
Error handling

rudehn:
Some general error handling for issue commands

### DIFF
--- a/cron/poll_read_issue_comments.py
+++ b/cron/poll_read_issue_comments.py
@@ -255,7 +255,6 @@ def is_command(comment):
     comment = re.sub('\s+', ' ', comment)
     parsed_comment = list(map(lambda x: x.lower(), comment.split(' ')))
     is_command = parsed_comment[0] in COMMAND_LIST  # Command will be at front
-    print("IS_COMMAND" + str(is_command))
     return is_command
 
 
--- a/cron/poll_read_issue_comments.py
+++ b/cron/poll_read_issue_comments.py
@@ -4,6 +4,7 @@ import json
 import os
 import re
 from os.path import join, abspath, dirname
+from requests.exceptions import HTTPError
 
 import settings
 import github_api as gh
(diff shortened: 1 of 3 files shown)


cron/poll_read_issue_comments.py
```diff
@@ -116,8 +117,12 @@ def insert_or_update(api, comment_id, issue_id, comment_txt):
...
 
-    seconds_remaining = gh.issues.voting_window_remaining_seconds(api, settings.URN, comment_id,
-                                                                  voting_window)
+    try:
+        seconds_remaining = gh.issues.voting_window_remaining_seconds(api, settings.URN, comment_id,
+                                                                      voting_window)
+    except HTTPError:
+        # Possible comment deleted
+        pass  # handle later
     seconds_remaining = max(0, seconds_remaining)  # No negative time
```
PlasmaPower: If an exception happens above, it will crash here. Also handle the exception with `__log.exception` please.

cron/poll_read_issue_comments.py
```diff
@@ -301,13 +311,17 @@ def poll_read_issue_comments(api):
...
+            # General exception catching
+            try:
+                # Get info and store in db
+                issue_id = issue_comment["issue_id"]
+                global_comment_id = str(issue_comment["global_comment_id"])
+                comment_text = issue_comment["comment_text"]
+                if is_command(comment_text):
+                    insert_or_update(api, global_comment_id, issue_id, comment_text)
+            except Exception as e:
+                __log.exception("Some exception occurred")
```
PlasmaPower, line 331: Please use a better error message, and just do `except:` instead of `except Exception as e:`.

cron/poll_read_issue_comments.py
```diff
@@ -192,8 +197,13 @@ def update_command_ran(api, comment_id, text):
...
     votes = {}
-    for voter, vote in gh.voting.get_comment_reaction_votes(api, urn, comment_id):
-        votes[voter] = vote
+
+    try:
+        for voter, vote in gh.voting.get_comment_reaction_votes(api, urn, comment_id):
+            votes[voter] = vote
+    except HTTPError:
+        # Command possibly deleted
+        pass  # Handle this later
```
PlasmaPower: Handle this with `__log.exception`

PlasmaPower (approved):

PlasmaPower (changes requested):
Oops meant to suggest changes, but I clicked approve.

cron/poll_read_issue_comments.py
rudehn: `log.exception` already prints the stack trace. I have no idea what exception would occur there, but I don't want to fatally crash the code

rudehn: Also I prefer to be explicit with what I'm catching, even if it's the general exception

PR #477, opened 2017-06-02
Better logging

mark-i-m:
- Reduce default logging level to INFO, rather than DEBUG
- Send log level ERROR or higher to stderr and everything else to stdout

### DIFF
--- a/chaos.py
+++ b/chaos.py
@@ -30,8 +30,7 @@ import encryption  # noqa: F401
 
 class LessThanFilter(logging.Filter):
     """
-    Source: https://stackoverflow.com/questions/2302315/how-can-info-and-
-    debug-logging-message-be-sent-to-stdout-and-higher-level-messag
+    Source: https://stackoverflow.com/questions/2302315
     """
     def __init__(self, exclusive_maximum, name=""):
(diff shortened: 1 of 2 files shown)


mark-i-m:
@PlasmaPower Meritocracy?

chaos.py
```diff
@@ -28,10 +28,36 @@
 import encryption  # noqa: F401
 
 
+class LessThanFilter(logging.Filter):
+    """
+    Source: https://stackoverflow.com/questions/2302315/how-can-info-and-
+    debug-logging-message-be-sent-to-stdout-and-higher-level-messag
```
PlasmaPower: Change this to just https://stackoverflow.com/questions/2302315 to avoid wrapping a link

PlasmaPower (approved):

PlasmaPower:
Sorry I confused this with #475. Reviewed!

chaos.py
mark-i-m: Oh, I didn't know you could do that. Great!

mark-i-m:
@PlasmaPower I made the fix you proposed...

PlasmaPower (approved):

PR #478, opened 2017-06-02
Fix rescinding reviews

PlasmaPower:
Previously, getting reviews would only return positive reviews. That's a problem! As noticed in #475.

### DIFF
--- a/github_api/voting.py
+++ b/github_api/voting.py
@@ -104,11 +104,10 @@ def get_pr_review_reactions(api, urn, pr):
     """ https://help.github.com/articles/about-pull-request-reviews/ """
     for review in prs.get_pr_reviews(api, urn, pr["number"]):
         state = review["state"]
-        if state in ("APPROVED", "DISMISSED"):
-            user = review["user"]["login"]
-            is_current = review["commit_id"] == pr["head"]["sha"]
-            vote = parse_review_for_vote(state)
-            yield user, is_current, vote
+        user = review["user"]["login"]
+        is_current = review["commit_id"] == pr["head"]["sha"]
+        vote = parse_review_for_vote(state)
+        yield user, is_current, vote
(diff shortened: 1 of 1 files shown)


mark-i-m (approved):

PlasmaPower:
Meritocracy review? @mark-i-m looks like you're barely not in the meritocracy :slightly_frowning_face:

mark-i-m:
Yep, I noticed that too... I get all the fun, no responsibilities :laughing:

andrewda (approved):

andrewda (approved):

andrewda:
Uhh the bot's on an hour long cooldown right now...

PlasmaPower:
I'm assuming you mean API cooldown. How often does the API limit refresh? Could we ask GitHub for extra requests?

andrewda:
Oh I see the issue – GitHub gave us a completely inaccurate reset time so it calculated the cooldowns wrong. Then GitHub told us we actually had 1 hour until the next reset so the bot has 18 requests left for the next 50 minutes.

andrewda:
Yea we could probably ask GitHub for a higher limit. I'm guessing they'll be pretty lenient as we haven't spammed the API and are pretty responsible with our API cooldowns.

hongaar:
@andrewda @PlasmaPower I was thinking about this yesterday. I think when we ask GH for more requests, they might ask us to implement [conditional requests](https://developer.github.com/v3/#conditional-requests) first. We might want to simply mirror the entire issue/pr list into our own DB, which should be much faster and durable. We could roll our own, or use [gousiosg/github-mirror](https://github.com/gousiosg/github-mirror) or [piwik/github-issues-mirror](https://github.com/piwik/github-issues-mirror) (unmaintained).

PR #479, opened 2017-06-02
Add roadmap (a link to issue Feature Requests and Ideas for Expansion wiki page from the homepage)

anythingbot:
Add a link to issue #473 to the homepage, [roadmap](https://github.com/chaosbot/Chaos/issues/473).

Issue #473, opened 2017-06-02
We are not creating chaos (Plan for rise of terminator)


ipranjal:
While main motive of chaosbot is to build something cool out of chaos, but most of the PR is based on changing or updating voting systems won't it be cool if we start to provide some intelligence to chaos bot.
 
*the terminator is coming* 

My roadmap :
1) Intigrate chatterbot to chaos bot (https://github.com/gunthercox/ChatterBot)
2) make a front end for people to interact with chatter-chaos bot 
3) make a twitter Handel where chaos bot will tweet once in a week some random stuff 

Goal : make a super machine developed by humans out of chaos 

More chaos can be added to chatterbot layer and chaos bot can be made to learn intresting things 

Rules :
1) if you are adding any feature listed on roadmap plz reference this issue so that everyone else can know 
2) If you like the idea upvote the issue so that I can get some motivation to work on it 😁

Let's stop working on democracy and meritocracy and actually create some chaos and cool stuff

o3LL:
Inb4 nazibot

mark-i-m:
@physcocode Feel free to submit a PR! I look forward to seeing what you come up with :smile:

anythingbot:
If you want, you could submit a PR that puts a link to this issue on the chaosthebot.com homepage with the text "roadmap"

chaosbot:
/vote close 

This issue hasn't been active for a while. To keep it open, react with :-1:

chaosbot:
> /vote close This issue hasn't been active for a while. To keep it open, react with :-1:

Time remaining: 2:10 - Vote status: passing :white_check_mark:

chaosbot:
> /vote close This issue hasn't been active for a while. To keep it open, react with :-1:

Vote Failed

chaosbot:
/vote close 

This issue hasn't been active for a while. To keep it open, react with :-1:

chaosbot:
> /vote close This issue hasn't been active for a while. To keep it open, react with :-1:

Command Ran

PR #480, opened 2017-06-02
Adding one more card to chaosbot.com

ipranjal:
This card can be used to display interesting concepts laid down for chaos bot related to disscussion on  #479 and tangentially related to #473

### DIFF
--- a/server/static/js/chaos.js
+++ b/server/static/js/chaos.js
@@ -145,7 +145,7 @@ document.getElementById("main").onclick = () => {
         document.getElementById("main").classList.add("active");
         document.getElementById("voters").classList.add("inactive");
         document.getElementById("voters").classList.remove("active");
-         document.getElementById("future").classList.add("inactive");
+        document.getElementById("future").classList.add("inactive");
         document.getElementById("future").classList.remove("active");
         document.getElementById("main").classList.remove("inactive");
     }
--- a/server/static/js/chaos.js
+++ b/server/static/js/chaos.js
@@ -126,7 +126,18 @@ document.getElementById("voters").onclick = () => {
     document.getElementById("voters").classList.add("active");
     document.getElementById("main").classList.add("inactive");
     document.getElementById("main").classList.remove("active");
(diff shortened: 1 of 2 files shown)


server/static/js/chaos.js
```diff
@@ -126,14 +126,27 @@ document.getElementById("voters").onclick = () => {
...
+    document.getElementById("main").classList.remove("active");
+    document.getElementById("future").classList.remove("inactive");
+};
+
 document.getElementById("main").onclick = () => {
     if (document.getElementById("main").classList.contains("inactive")) {
         document.getElementById("main").classList.add("active");
         document.getElementById("voters").classList.add("inactive");
         document.getElementById("voters").classList.remove("active");
+         document.getElementById("future").classList.add("inactive");
```
andrewda: fix spacing

andrewda (changes requested):

server/static/js/chaos.js
ipranjal: yep was just doing that

andrewda (approved):

ipranjal:
@PlasmaPower you dont seem much happy about this :P

PlasmaPower:
@physcocode no, the point of a wiki is sorta that it's easier to change (you don't need a PR). This defeats the point.

andrewda:
It'd be cool if this could either a) just link to the feature request wiki page, or b) scrape the wiki page and display the content.

ipranjal:
@PlasmaPower @andrewda just wanna give democracy over what people actually wanted to display on the list of future concept . people could down vote what they don't want to add in list

mdcfe:
👎 for two reasons:
* Static content on the homepage - as @PlasmaPower said, we can use the wiki
* The cards on the homepage are already clunky and don't work well on mobile. If we can improve how they work and look (for example no animations on mobile, maximum height to prevent scrollbars) then I'd be fine with adding more cards.

ipranjal:
Okay I am down voting this the card also don't work ok safari it's frikin broken

PR #481, opened 2017-06-02
Error handling

rudehn:
Some general error handling for issue commands.

Issue commands won't work without this.

### DIFF
--- a/github_api/comments.py
+++ b/github_api/comments.py
@@ -46,10 +46,7 @@ def get_reactions_for_comment(api, urn, comment_id):
     params = {"per_page": settings.DEFAULT_PAGINATION}
     reactions = []
 
-    try:
-        reactions = api("get", path, params=params)
-    except:
-        pass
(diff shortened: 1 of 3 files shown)


Swizz:
:+1: on principle

mdcfe (approved):

github_api/comments.py
```diff
@@ -46,10 +46,7 @@ def get_reactions_for_comment(api, urn, comment_id):
     params = {"per_page": settings.DEFAULT_PAGINATION}
     reactions = []
 
-    try:
-        reactions = api("get", path, params=params)
-    except:
-        pass
+    reactions = api("get", path, params=params)
```
PlasmaPower, line 49: Why remove error handling here?

rudehn: If this fails, it's most likely due to a comment was deleted. If a comment was deleted, it shouldn't fail silently but rather should be handled

PR #484, opened 2017-06-02
fix invalid syntax

rudehn:
This is invalid syntax and is causing a revert of all new PRs. Fixing this should allow all new PRs to no longer be reverted.

### DIFF
--- a/lib/db.py
+++ b/lib/db.py
@@ -4,7 +4,6 @@
 from importlib import import_module
 
 import settings
-from lib.db_adapters.interface import AbstractAdapter
 
 
 class DB(object):
--- a/lib/db.py
+++ b/lib/db.py
@@ -20,7 +20,7 @@ class DB(object):
         return DB.instance
 
     @staticmethod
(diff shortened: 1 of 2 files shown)


Ayyko:
https://www.python.org/dev/peps/pep-3107/#return-values 
What is getting mad about this?

rudehn:
```
SyntaxError: invalid syntax
  File "./lib/db.py", line 23
    def create_adapter(adapter_name, config) -> AbstractAdapter:
                                             ^
SyntaxError: invalid syntax
```

```
File "./lib/db_adapters/interface.py", line 7
    class AbstractAdapter(object, metaclass=abc.ABCMeta):
                                           ^
SyntaxError: invalid syntax
```

If it doesn't like it, it doesn't like it

mdcfe:
```
$ flake8
./lib/db.py:7:1: F401 'lib.db_adapters.interface.AbstractAdapter' imported but unused
```

Ayyko:
¯\\\_(ツ)_/¯ fair enough

PlasmaPower (approved):

PR #486, opened 2017-06-02
Create and update labels improved, close #476

### LINKED ISSUE #476: GitHub API: 422 unprocessable entity
> 422 Client Error: Unprocessable Entity for url: https://api.github.com/repos/chaosbot/Chaos/labels

http://chaosthebot.com:8081/chaos-stderr.log

We had this issue with status updates, but now it's returned in the form of label creation. Ideas?

Swizz:
Improve label creation

- Attempt to create label
- If label exists update it regardless of changes
- Updating name causes new label creation for compatibility purpose

Use `create_label` all time if this is possible.

NB : [I have tested this PR on my local instance this time !](https://github.com/Swizz/Chaos/labels)

### DIFF
--- a/github_api/repos.py
+++ b/github_api/repos.py
@@ -53,6 +53,8 @@ def create_label(api, urn, name, color="ededed"):
     except HTTPError as e:
         if e.response.status_code == 422:
             update_label(api, urn, name, color)
+        else:
+            __log.exception("couldn't create label")
     return resp
 
 
@@ -66,5 +68,5 @@ def update_label(api, urn, name, color="ededed"):
     try:
         resp = api("patch", "/repos/{urn}/labels/{name}".format(urn=urn, name=name), json=data)
     except HTTPError:
-        pass
+        __log.exception("couldn't update label")
     return resp
--- a/github_api/repos.py
+++ b/github_api/repos.py
@@ -1,6 +1,7 @@
 import arrow
 import settings
 import logging
+from requests import HTTPError
 
 __log = logging.getLogger("github_api.repos")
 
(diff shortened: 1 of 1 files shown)


github_api/repos.py
```diff
@@ -49,7 +50,21 @@ def create_label(api, urn, name, color="ededed"):
...
+    """ update an issue label for the repository """
+    data = {
+        "name": name,
+        "color": color
+    }
+    resp = None
+    try:
+        resp = api("patch", "/repos/{urn}/labels/{name}".format(urn=urn, name=name), json=data)
+    except HTTPError:
+        pass
```
mdcfe: Shouldn't we be at least logging this error? If the label couldn't be updated then there was probably a reason.

github_api/repos.py
```diff
@@ -49,7 +50,21 @@ def create_label(api, urn, name, color="ededed"):
     resp = None
     try:
         resp = api("post", "/repos/{urn}/labels".format(urn=urn), json=data)
-    except:
-        __log.exception("couldn't create label")
+    except HTTPError as e:
+        if e.response.status_code == 422:
+            update_label(api, urn, name, color)
```
mdcfe, line 55: Same here - if the code isn't a 422, shouldn't we still log it?

mdcfe (changes requested):

github_api/repos.py
Swizz: Got it

PlasmaPower (approved):

mdcfe (approved):

PR #488, opened 2017-06-02
Switch to a hug-based web API with uWSGI

mdcfe:
This will make life much easier when we move from dumping JSON files in the server root to using a database (MySQL/SQLite).

I'm not sure what changes will need to be made on the real ChaosBot server to get Nginx to forward uWSGI properly, but I've modified the Docker `start_services.sh` script to link the new config properly.

(Note: ~~69d7aff says "incomplete" because currently, `/` serves the hug help API and you need to visit `/index.html` to show the index.~~ Fixed in 012513b)


webserver.py
```diff
@@ -0,0 +1,30 @@
...
+        __log.exception("Failed to read voters!")
+
+
+@hug.get("/meritocracy", examples=["amount=5", "amount=0"])
+def get_meritocracy(amount: hug.types.number = 20):
+    try:
+        with open("/root/workspace/Chaos/server/meritocracy.json", "r") as f:
+            meritocracy = json.load(f)
+        if amount > 0:
+            meritocracy = meritocracy[:amount]
```
rudehn: is meritocracy defined here?

webserver.py
```diff
@@ -0,0 +1,30 @@
...
+__log = logging.getLogger("webserver")
+
+
+@hug.get("/voters", examples=["amount=20", "amount=0"])
+def get_voters(amount: hug.types.number = 20):
+    try:
+        with open("/root/workspace/Chaos/server/voters.json", "r") as f:
+            voters = sorted(json.load(f).items(), key=lambda x: x[1])
+        if amount > 0:
+            voters = voters[:amount]
```
rudehn: is voters defined here?

mdcfe: Yes, it is. If the sorted call returns null for some reason, it will be caught.

webserver.py
mdcfe: Same as before

webserver.py
rudehn: I mean is voters initially scope limited to the `with` block?

hongaar:
Have no experience with hug or uwsgi but looks incredible, so :+1:

webserver.py
mdcfe: I've tested it and that doesn't appear to be the case.

mdcfe:
@hongaar I don't have much experience with them either 😛

webserver.py
PlasmaPower: @rudehn python scopes are classes and functions only.

webserver.py
```diff
@@ -0,0 +1,30 @@
...
+import logging
+
+
+__log = logging.getLogger("webserver")
+
+
+@hug.get("/voters", examples=["amount=20", "amount=0"])
+def get_voters(amount: hug.types.number = 20):
+    try:
+        with open("/root/workspace/Chaos/server/voters.json", "r") as f:
```
PlasmaPower: Cache this?

webserver.py
```diff
@@ -0,0 +1,30 @@
...
+            voters = sorted(json.load(f).items(), key=lambda x: x[1])
+        if amount > 0:
+            voters = voters[:amount]
+        return {x[0]: x[1] for x in voters}
+    except:
+        __log.exception("Failed to read voters!")
+
+
+@hug.get("/meritocracy", examples=["amount=5", "amount=0"])
+def get_meritocracy(amount: hug.types.number = 20):
```
PlasmaPower, line 30: Do we really need a maximum amount for this?

Swizz:
@hongaar If you take a look at the gitter chat, we choose hug and moon for the web api/site for ease.
We heavily decided to choose simplicity and minimalism first.

@md678685 Do you need me on some stuff ?

mdcfe:
@Swizz If you want to add anything else (or you're particularly experienced with Nginx 😉) then that'd be great!

Swizz:
Front end dev, know nothing about Nginx

mdcfe:
Fair enough, I'll try and sort the index issue out

mdcfe:
I think this is ready for merging - could we get a meritocracy review? @PlasmaPower

PlasmaPower (changes requested):
This will currently break the top voters card, as you're requesting the first 20 voters, not the top 20 voters.

mdcfe:
@PlasmaPower The voters are now sorted in descending (highest to lowest) order before it gets limited by the amount.

PlasmaPower (approved):

Issue #180, opened 2017-05-24
Put reddit password and secret key in inaccessable file


yet-another-account:
Just like the github token

amoffat:
do you mean the secret key that would be generated by https://github.com/chaosbot/chaos/pull/141 ?

qgustavor:
The take 5 was merged #227: a pure RSA implementation (key length: 2048, hash: SHA1). The maximum size of encrypted content is 214 bytes (based on [those calculations](https://crypto.stackexchange.com/a/42100)). If someone wants to store anything larger than that implement AES ([doing it wrong](https://people.eecs.berkeley.edu/~daw/teaching/cs261-f12/misc/if.html)).

mpnordland:
Just a note, the account in the file appears to have been stolen, I've created a new account, /u/chaosthebotreborn with new info and I'm working on a pr with a bot and an encrypted config file.

chaosbot:
This issue hasn't been active for a while.To keep it open, react with :-1: on the `vote close` post.

chaosbot:
> /vote close

Vote failed

chaosbot:
/vote close 

This issue hasn't been active for a while.To keep it open, react with :-1:

chaosbot:
> /vote close This issue hasn't been active for a while.To keep it open, react with :-1:

Command Ran

chaosbot:
:no_entry: The issue has been closed after a vote.

chaosbot:
> /vote close

Vote Failed

chaosbot:
> /vote close

Vote Failed

chaosbot:
> /vote close

Vote Failed

chaosbot:
> /vote close

Vote Failed

chaosbot:
> /vote close

Vote Failed

chaosbot:
> /vote close

Vote Failed

chaosbot:
> /vote close

Vote Failed

chaosbot:
> /vote close

Vote Failed

chaosbot:
> /vote close

Vote Failed

chaosbot:
> /vote close

Vote Failed

chaosbot:
> /vote close

Vote Failed

chaosbot:
> /vote close

Vote Failed

chaosbot:
> /vote close

Vote Failed

chaosbot:
> /vote close

Vote Failed

[...]

PR #492, opened 2017-06-03
Improved reddit bot and symmetric encryption

mpnordland:
This is an omnibus pr. I needed proper file encryption for the redditbot config file.

# The Reddit Bot
Watches for comments in /r/chaosthebot with the text "hey chaosbot" and replies with "Hey <username>!" Should only respond once, if it responds twice, pm /u/mpnordland.

# Symmetric Key Encryption
The public key crypto available wouldn't be secure for a file larger than the key. So, I introduce a symmetric key management and file encryption system. I have also modified encryption.py to allow for encrypting things as well. Using the KeyManager class from symmetric_keys.py, you can transparently handle encrypted keys from symmetric_keys.json. KeyManager handles the public key crypto for encrypting and decrypting the symmetric keys. encrypt_config_file.py uses KeyManager to encrypt the file of your choosing with a freshly generated symmetric key. The new key is then encrypted with the server's pubkey and stored under a name of your choosing in symmetric_keys.json.


PlasmaPower:
> The public key crypto available wouldn't be secure for a file larger than the key.

Umm... why do you think that? Assuming the public key crypto is done correctly, it's perfectly secure. It might be slightly less secure than symmetric crypto, but both are practically impossible to crack with a significantly large key size (1024 bits).

PlasmaPower:
So let me get this straight. This PR's process is:

- hardcoded encrypted key
- decrypt that key with the current public key system
- use that key to decrypt the hardcoded encrypted reddit token

Why? I'm pretty sure the existing public key system is fine by itself.

chaos.py
```diff
@@ -118,7 +128,8 @@ def check_for_prev_crash(api, log):
             # Currently, I'm just reading the last 200 lines... which I think
             # ought to be enough, but if anyone has a better way to do this,
             # please do it.
-            dump = subprocess.check_output(["tail", "-n", "200", settings.CHAOSBOT_STDERR_LOG])
+            dump = subprocess.check_output(
+                ["tail", "-n", "200", settings.CHAOSBOT_STDERR_LOG])
```
PlasmaPower, line 132: Why the whitespace changes?

PlasmaPower:
The reddit bot itself looks good to me. Still really confused about the extra crypto though.

mpnordland:
@PlasmaPower I saw in this issue that it would only work for ~200 bytes https://github.com/chaosbot/Chaos/issues/180 . Without revealing too much, I believed I would be over that amount, therefore I followed the recommendation in that comment. The cryptography library's Fernet Class used AES to encrypt, plus was easy to use so I went with that.

PlasmaPower:
Oh I thought the crypto library would automatically split it into blocks. I might be wrong. I'm also surprised the secret is more than 200 bytes, but that's fine.

PlasmaPower (approved):

PlasmaPower:
@anythingbot no encryption is used here so the author doesn't have to contact the sysadmin. We'd prefer to keep this project self-maintaining.

mpnordland:
Problem: how will the bot inform Reddit of this new password?

---- On Sat, 03 Jun 2017 12:40:26 -0400 notifications@github.com wrote ----

@PlasmaPower re: the need for encryption to store the config file

If the bot auto-generates a reddit password, then it can combine a redditbot template config file with the auto-generated password and stick he result in a file that is in .gitignore.

I don't think encryption is actually needed here if the bot generates the password and keeps it hidden.

I don't think encryption makes this more secure: either one of us will have the private key, or else the bot will have to create a private key and keep it hidden (in either case, either the reddit config or the private key will have to be kept hidden somehow).

PlasmaPower:
@anythingbot then the bot would have to create the account too, which is (for obvious reasons) not supported by the API, and requires solving a captcha.

PlasmaPower:
@mpnordland exactly

mpnordland:
And I thought my solution was complicated. Are sure you're not trolling?

---- On Sat, 03 Jun 2017 12:58:19 -0400 notifications@github.com wrote ----

@PlasmaPower @mpnordland If all you want is a way for the bot to know the correct password without checking it into git, then you can put up a /redditbotconfig webpage that lets a web visitor submit the reddit password over HTTP (I know, this isn't secure against an attacker listening to TCP packets, but it is more secure than checking the reddit password into git and publishing it on github). The bot will check that the password works and remember it (store the password in a file whose name appears in .gitignore). After this is done, the bot disables the /redditbotconfig web page since it won't be used again.

This only has to be done once, since the bot will have the password in a local file and won't put up the /redditbotconfig page if this file exists.

If you want even more protection (against an attacker listening to TCP packets) then you can have the bot generate a pair of asymmetric encryption keys, put the private key in a file in .gitignore, publish the public key in the /redditbotconfig page, and ask the user to encrypt the reddit password with the public key before submitting it.

PlasmaPower:
I always wonder with anythingbot.

andrewda:
I mean... It would technically work...

PlasmaPower:
So would the solution we have in front of us, already implemented.

mdcfe:
> If you want even more protection (against an attacker listening to TCP packets) then you can have the bot generate a pair of asymmetric encryption keys, put the private key in a file in .gitignore, publish the public key in the /redditbotconfig page, and ask the user to encrypt the reddit password with the public key before submitting it.

But... we already have a keypair specifically for this...?

PlasmaPower:
Reopen this PR with an explanation of the encryption and it might pass.

Issue #401, opened 2017-05-30
Get bug fixes in quicker


rudehn:
It would be nice to have some solution to get bug fixes (especially fatal ones) without having to wait through the full voting window.

PlasmaPower:
Any ideas? Maybe add an additional threshold for an instamerge, even before the extended voting window.

rudehn:
No ideas yet..

rudehn:
So I'm starting to get some kind of idea..

Here are the steps I'm thinking

1. Chaosbot detects the error. Without this we have no idea we need to fix it.
2. Chaosbot submits an issue with the error, and asks if we want a PR with a bug fix.
3. Use a /fix #PR command syntax, or something similar

There would have to be an adjustment to the voting thresholds I'm thinking. It would probably need a meritocracy vote.

Can you think of how this method could be exploited?

PlasmaPower:
That could work, but we might need punishments for abusing `/fix`.

rudehn:
I suppose one question is do we want the PR owner to be the only one able to `/fix`? Do we want anybody? Or anybody but PR owner?

We would probably need a 30 min window at least just to be safe.

mdcfe:
This could facilitate sneaking in malicious changes, though - especially if it's just a small change that might get missed with the shorter window and spotted once it has passed.

rudehn:
Yeah there's the risk of that. That's why I was trying to see if there was some approach that could limit or even prevent that from occurring.

PlasmaPower:
@md678685 the meritocracy might help, but yes it still remains a problem

mark-i-m:
> Maybe add an additional threshold for an instamerge, even before the extended voting window.

Along the same lines, we could make it so that if a majority of the meritocrats submit an approving review, it gets instantly merged. That seems like a pretty high bar, though... We could make it something like 5 reviews, but that also seems like a pretty high bar...

mark-i-m:
Ok, here is a new proposal: The author of the PR includes `/expedite` or something in a post on the PR. If there are 5 approving current reviews the PR gets merged.

PlasmaPower:
@mark-i-m I think `/expedite` would get overused.

mark-i-m:
I suspect that if the bar is significantly high, people won't have high hopes to overuse this. Perhaps we could make it a majority of the meritocracy... a meritjority

andrewda:
[...]

PR #493, opened 2017-06-03
[WIP] brainstorming for /vote fast + prototype [WILL NOT BE UN-WIP-ED]

mark-i-m:
Start work for /vote fast

My proposal is [here](https://github.com/chaosbot/Chaos/issues/401#issuecomment-305939080) in #401 .

~This PR is not intended to implement any feature, but to implement some needed functions/methods/utilities that might be needed, and to just plan in general.~

~The actual feature will probably wait for @rudehn's work on database stuff.~

Thoughts and comments appreciated...

EDIT: I added some code to implement this feature now that @rudehn's work is in... I will probably not un-WIP this. Rather, when it looks good, I will create a new PR, so as not to take advantage of those who already voted.


mark-i-m:
Useful links:

https://developer.github.com/v3/issues/comments/#list-comments-on-an-issue

https://developer.github.com/v3/pulls/#get-a-single-pull-request

mdcfe:
@anythingbot The proposal for the feature is... wait, he already linked it:
> My proposal is [here](https://github.com/chaosbot/Chaos/issues/401#issuecomment-305939080) in #401.

This is marked WIP - it's clearly not finished yet, and mark even said that this is only the beginning of the work for it.

rudehn:
@mark-i-m you will need to change the `COMMANDS_LIST` variable and add the `fast` subcommand to it. Command syntax is checked against whatever is in the commands_list.

Ajedi32:
Wait, does chaosbot avoid automatically merging [WIP] PRs? I didn't realize that was something it did.

rudehn:
Yep. If it has `WIP` it will not be merged

rudehn:
DB is in, this can be finished

mark-i-m:
@rudehn Can you confirm that I am not messing up the database stuff?

rudehn:
Walk me through what you are trying to do. What does `/vote fast` do?

mark-i-m:
1) When a new PR is created, a meritocrat has 5 min to post a `/vote fast`
2) If the comment is edited it is ignored
3) If (1) and (2) hold, then a marker is placed on the PR in the db... Specifically, the Issue table is augmented with an "expedited" column.
4) During the PR polling, if the PR (which is an issue w.r.t. github api IIUC) has an "expedited" flag set in its record in the db, then the voting window is halfed.

cron/poll_pull_requests.py
```diff
@@ -73,6 +73,16 @@ def poll_pull_requests(api):
             # the PR is mitigated or the threshold is not reached ?
             if variance >= threshold or not is_approved:
                 voting_window = gh.voting.get_extended_voting_window(api, settings.URN)
+
+                # read the db to see if this PR is expedited by /vote fast...
+                try:
+                    issue = Issue.get(issue_id=pr["id"])
+                    if issue.expedited: # expedite
+                        voting_window /= 2
```
PlasmaPower: This code needs to be moved out of the if statement for the extended voting window. Currently this just halves the extended voting window.

cron/poll_read_issue_comments.py
```diff
@@ -145,6 +151,33 @@ def handle_vote_command(api, command, issue_id, comment_id, votes):
...
+            # The post should not have been updated
+            if updated_at != created_at:
+                return
+
+            # The comment poster must be in the meritocracy
+            meritocracy = {}
+            with open('server/meritocracy.json', 'r') as mfp:
+                fs = fp.read()
+                if fs:
+                    meritocracy = json.loads(fs)
```
PlasmaPower: This seems a bit flaky. How about defining a function `get_meritocracy` with the current code for generating the meritocracy?

cron/poll_pull_requests.py
```diff
@@ -73,6 +73,16 @@ def poll_pull_requests(api):
...
+
+                except Issue.DoesNotExist:
+                    pass # not expedited
```
PlasmaPower: Also, add an `except: log.exception("Failed to get expedited status")`

PlasmaPower (changes requested):
There are also a bunch of flake8 warnings, and some of them are critical (e.g. you forgot to import `json` in `cron/poll_read_issue_comments.py`).

PlasmaPower:
@mark-i-m Don't you also want to require additional meritocracy reviews? Also, should the author of the PR be able to `/vote fast` it?

cron/poll_read_issue_comments.py
mdcfe: `server/meritocracy.json` is temporary - it shouldn't ideally be read from because it's only there to fill the gap before the database is ready and we can use that instead.

dbpokorny:
@mark-i-m this can be done with a mysql database, but this isn't necessary; it can also be done with a github label. That way the bot can check for labels on the PR to see if the "expedited" label is present. If so, the bot cuts the voting window in half.

PlasmaPower:
@dbpokorny That's a great idea! I'm not sure it'll be necessary though, since we do have a DB and an implementation using it.

cron/poll_pull_requests.py
mark-i-m: @PlasmaPower This is not necessarily a failure, though... It just means nobody has posted `/vote fast` on this PR. In fact, we expect that to be the common case, so these would flood the logs.

cron/poll_read_issue_comments.py
mark-i-m: @md678685 What should I do instead? Would it be ok to define `get_meritocracy` as @PlasmaPower said and simply re-implement it when the db is ready?

mark-i-m:
Adding the label does sound like a good way of visually differentiating fast PRs on the PR list, though...

cron/poll_pull_requests.py
PlasmaPower: I'm not sure you understand. I'm saying keep the current error handler, but add another general one in case a different exception happens.

cron/poll_read_issue_comments.py
PlasmaPower: I don't think it would need re-implemented when the DB is ready. Sure, the DB could be used, but there's no point in that since this code and the main loop (PRs) run in the same process (so they share the same memoization cache).

dbpokorny:
@PlasmaPower My whole take on expedited PRs is that it effectively turns the chaosbot into a bicameral legislature; so I think either method is fine. Part of the issue here is that using a mysql database doesn't reveal the status of the PR using the github web interface; you'll still have to add the github label in order to signal the status of the PR to github users. So, the thinking goes, if github is going to remember the status, then there isn't any need (on this project) for additional mysql configuration.

I'm under the impression that the only data this database would store is expedited status for each PR. Is that correct?

PlasmaPower:
@dbpokorny That is a good point in favor of a label. I guess we could add a label, but pull the info from the DB. And yes, I think you're correct about the DB model.

cron/poll_pull_requests.py
mark-i-m: Ah, my bad... fixed

dbpokorny:
@PlasmaPower pull what info from the database? I don't understand what that is

cron/poll_read_issue_comments.py
```diff
@@ -36,16 +36,31 @@ def get_seconds_remaining(api, comment_id):
...
+    # get more info on the issue
+    gh_issue = gh.issue.open_issue(api, settings.URN, number)
+
+    # get user from db
+    user, _ = User.get_or_create(user_id=gh_issue["user"]["id"],
+                                 defaults={"login": gh_issue["user"]["login"]})
+
+    # db insert
+    issue, _ = Issue.get_or_create(issue_id=issue_id,
+                                   number=number,
```
mark-i-m: Does anyone know what the difference between these is? Can I drop one of them?

mark-i-m:
Ok, I think I have code representing roughly what has been discussed above. I really doubt it runs in its current state. Mostly, I would appreciate feedback from reviews about what should be changed. After that, I will fix up broken things and submit a new PR. Think of this PR as pseudocode for the real one...

mark-i-m:
Will be back tomorrow :zzz:

mark-i-m:
> Also, should the author of the PR be able to /vote fast it?

Sorry, I didn't see this earlier. I don't think so. If it is something that should be `/vote fast`ed, then the mertiocracy should be aware of it. Otherwise, I think it will just get abused.

PlasmaPower:
First issues to fix should be Travis. There are some undefined names and such.

cron/poll_read_issue_comments.py
rudehn: I would recommend using the `defaults` param here for anything not in the primary key. We were having issues using this syntax before

cron/poll_read_issue_comments.py
```diff
@@ -35,13 +36,31 @@ def get_seconds_remaining(api, comment_id):
...
+
+
 def insert_or_update(api, cmd_obj):
     # Find the comment, or create it if it doesn't exit
     comment_id = cmd_obj["global_comment_id"]
-    issue, _ = Issue.get_or_create(issue_id=cmd_obj["issue_id"])
     user, _ = User.get_or_create(user_id=cmd_obj["user"]["id"],
                                  defaults={"login": cmd_obj["user"]["login"]})
-
+    issue = insert_or_update_issue(api, cmd_obj["issue_id"], cmd_obj["number"])
```
rudehn, line 66: where are you getting `cmd_obj["number"]` from? I don't think this exists

cron/poll_read_issue_comments.py
```diff
@@ -145,6 +178,48 @@ def handle_vote_command(api, command, issue_id, comment_id, votes):
...
+            comment_poster = cmdmeta.comment.user
+            issue_poster = cmdmeta.issue.user
+            issue_created_at = cmdmeta.issue.created_at
+
+            # This must be a PR
+            if not cmdmeta.issue.is_pr:
+                return
+
+            # The post should not have been updated
+            if updated_at != created_at:
```
rudehn: maybe leave some kind of response indicating why the command didn't run? Makes debugging easier

lib/db/models.py
```diff
@@ -28,6 +28,11 @@ class Comment(BaseModel):
 
 class Issue(BaseModel):
     issue_id = pw.IntegerField(primary_key=True)
+    number = pw.IntegerField()
```
rudehn, line 31: what is this number?

cron/poll_read_issue_comments.py
```diff
@@ -145,6 +178,48 @@ def handle_vote_command(api, command, issue_id, comment_id, votes):
             gh.issues.close_issue(api, settings.URN, issue_id)
         elif sub_command == "reopen":
             gh.issues.open_issue(api, settings.URN, issue_id)
+        elif sub_command == "fast":
```
rudehn: maybe pull this out into it's own function?

cron/poll_read_issue_comments.py
mark-i-m: Better?

cron/poll_read_issue_comments.py
mark-i-m: I think it ought to be just returned from github, right: https://developer.github.com/v3/issues/#list-issues-for-a-repository?

Am I missing something?

cron/poll_read_issue_comments.py
mark-i-m: Hmm... yes, I see your point... but would this be prone to inducing spam?

lib/db/models.py
mark-i-m: The PR number as oppose to PR id? The thing is when github returns a PR, it returns a value called "id" which is not the PR number we see on github, and it returns a "number" which is the number we see on github. TBH, I am really confused about which to use where. I think we want to use issue/PR numbers, rather than ids, but the db models appear to use the ids. Any help/auditing on this point would be appreciated.

cron/poll_read_issue_comments.py
mark-i-m: Done

cron/poll_read_issue_comments.py
rudehn: Yep. Read comments.py / get_all_issue_comments. It doesn't return the entire structure. You might need to change that

lib/db/models.py
rudehn: The ids are used when making api calls. I think that's why we store them

cron/poll_read_issue_comments.py
mark-i-m: Ah, good catch. Fixed

lib/db/models.py
mark-i-m: Hmm... interesting. Most of the API I saw for issues and PRs take the issue/PR number, though?

cron/poll_read_issue_comments.py
```diff
@@ -135,7 +157,63 @@ def get_command_votes(api, urn, comment_id):
     return votes
 
 
-def handle_vote_command(api, command, issue_id, comment_id, votes):
+def get_meritocracy():
+    # FIXME: update when this is done by the db
```
PlasmaPower: I'm not sure the meritocracy will ever be stored in the DB. Instead, put [this code](https://github.com/chaosbot/Chaos/blob/e0e352896eddec03c01ba4e85b1263b83bb669f0/cron/poll_pull_requests.py#L40-L51) in the function.

PlasmaPower:
[This code](https://github.com/chaosbot/Chaos/blob/e0e352896eddec03c01ba4e85b1263b83bb669f0/cron/poll_pull_requests.py#L40-L51) should also be changed to use the `get_meritocracy` function (maybe move it into a more general file).

mark-i-m:
I would like to close this PR to clean up my dashboard. Does anyone have objections?

mdcfe:
As with the other issue, no objection here.

PR #495, opened 2017-06-03
Invert vote weight.

lurr:
Will probably totally not break something


PlasmaPower:
That would mean upvotes are downvotes, and downvotes are upvotes for anyone confused.

mark-i-m:
I love it!

But maybe we could enable it only on Tuesdays or something...

PR #500, opened 2017-06-03
Fix meritocracy mentions and extended voting window

PlasmaPower:
Now it must be running in production, and it will not mention the author of the PR (as they cannot review their own PR).

Edit: in addition, the PR must not be in the base voting window.

### DIFF
--- a/cron/poll_pull_requests.py
+++ b/cron/poll_pull_requests.py
@@ -79,16 +79,17 @@ CREATE TABLE IF NOT EXISTS meritocracy_mentioned (
             # the PR is mitigated or the threshold is not reached ?
             if variance >= threshold or not is_approved:
                 voting_window = gh.voting.get_extended_voting_window(api, settings.URN)
-                if vote_total >= threshold / 2:
+                if settings.IN_PRODUCTION and vote_total >= threshold / 2:
                     # check if we need to mention the meritocracy
                     try:
                         commit = pr["head"]["sha"]
                         if not db.query("SELECT * FROM meritocracy_mentioned WHERE commit_hash=?",
                                         (commit,)):
+                            meritocracy_mentions = meritocracy - {pr["user"]["login"].lower()}
                             db.query("INSERT INTO meritocracy_mentioned (commit_hash) VALUES (?)",
                                      (commit,))
                             gh.comments.leave_meritocracy_comment(api, settings.URN, pr["number"],
-                                                                  meritocracy)
+                                                                  meritocracy_mentions)
                     except:
                         __log.exception("Failed to process meritocracy mention")
 
--- a/cron/poll_pull_requests.py
+++ b/cron/poll_pull_requests.py
@@ -29,7 +29,7 @@ CREATE TABLE IF NOT EXISTS meritocracy_mentioned (
         __log.exception("Failed to create meritocracy mentioned DB table")
 
     # get voting window
-    voting_window = gh.voting.get_initial_voting_window()
+    base_voting_window = gh.voting.get_initial_voting_window()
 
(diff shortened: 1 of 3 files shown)


Swizz (approved):

cron/poll_pull_requests.py
```diff
@@ -79,16 +79,17 @@ def poll_pull_requests(api):
...
             if variance >= threshold or not is_approved:
                 voting_window = gh.voting.get_extended_voting_window(api, settings.URN)
-                if vote_total >= threshold / 2:
+                if settings.IN_PRODUCTION and vote_total >= threshold / 2:
                     # check if we need to mention the meritocracy
                     try:
                         commit = pr["head"]["sha"]
                         if not db.query("SELECT * FROM meritocracy_mentioned WHERE commit_hash=?",
                                         (commit,)):
+                            meritocracy_mentions = meritocracy - {pr["user"]["login"].lower()}
```
mark-i-m: Should we have the `.lower()` here?

PlasmaPower: Yeah, the meritocracy set is all lowercase.

mark-i-m: hmm... are you sure? The comment in #499 is not all lowercase... :confused:

PlasmaPower: GitHub automatically fixes mention cases. Try mentioning me like this: `@plasmapower`. You'll notice it turns into `@PlasmaPower`.

PlasmaPower:
Oops just realized the code for the extended voting window is still broken (as seen in #503). Fixing.

cron/poll_pull_requests.py
mark-i-m: @plasmapower

mark-i-m: Hmm... interesting :+1:

Swizz:
Chaosbot still stealing a Meritocratic chair

mark-i-m:
@chaosbot will be in the meritocracy for a while because it was the merger on all PRs before we switched to squashing...

mark-i-m:
@PlasmaPower while we're at it, you could filter out @chaosbot from the meritocracy review post.

PlasmaPower:
@mark-i-m Will do. BTW I figured out why the extended voting window was/is malfunctioning, will explain after fix.

cron/poll_pull_requests.py
```diff
@@ -76,30 +76,36 @@ def poll_pull_requests(api):
             threshold = gh.voting.get_approval_threshold(api, settings.URN)
             is_approved = vote_total >= threshold and meritocracy_satisfied
 
+            seconds_since_updated = gh.prs.seconds_since_updated(api, settings.URN, pr)
+
+            voting_window = base_voting_window
             # the PR is mitigated or the threshold is not reached ?
             if variance >= threshold or not is_approved:
                 voting_window = gh.voting.get_extended_voting_window(api, settings.URN)
```
PlasmaPower, line 84: Previously, this would update the global `voting_window`, as Python doesn't have any scope beyond functions and classes. That meant any PR processed after one with the extended voting window would also get the extended voting window.

PlasmaPower:
This now also saves an extra API call per PR check (it got the last updated time twice).

PlasmaPower:
Rip chaosbot again? http://chaosthebot.com:8081/chaos-stderr.log

That 404 seems to be common, not quite sure why it happens though.

PlasmaPower:
(this PR actually needs a meritocracy review - @andrewda?)

andrewda (approved):
👍

o3LL (approved):

PR #501, opened 2017-06-03
fix Patchset from requests

Swizz:
Getting patch raise no error but was in fact empty.

### DIFF
--- a/github_api/__init__.py
+++ b/github_api/__init__.py
@@ -101,6 +101,6 @@ class API(object):
         try:
             data = resp.json()
         except:
-            data = None
+            data = resp.text
 
         return data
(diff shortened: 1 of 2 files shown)


PlasmaPower (approved):
I looked through the code, and I don't think the return change will break anything.

Swizz:
@PlasmaPower This was introduced here https://github.com/chaosbot/Chaos/commit/5a2d401c22361d72922085cb25c2d4748a627c6c for hotfix purpose, no more use in code.

andrewda:
There are some bugs in chaosbot's review needed comments ;)

PlasmaPower:
@andrewda Yep, fixed in #500

PR #502, opened 2017-06-03
Add a label to PR regarding changes

Swizz:
Files and directories are under "part name".
The cron task will read the patch file from the PR and apply a label regarding changes.

labels on PRs

### DIFF
--- a/cron/poll_pull_requests.py
+++ b/cron/poll_pull_requests.py
@@ -5,6 +5,7 @@ import sys
 from os.path import join, abspath, dirname
 from lib.db import DB
 
+import lib
 import settings
 import github_api as gh
 
@@ -94,6 +95,17 @@ CREATE TABLE IF NOT EXISTS meritocracy_mentioned (
 
             # is our PR in voting window?
             in_window = gh.prs.is_pr_in_voting_window(api, pr, voting_window)
+
+            # Add the rights labels
+            __log.info("Adding PR %d labels regarding is patch", pr_num)
+
+            patched_files = gh.prs.get_patch(api, settings.URN, pr_num)
+            patched_parts = [lib.part_name(settings.SOURCE_PARTS_NAME, patched_file.path)
+                             for patched_file in patched_files]
+
+            patched_parts = list(set(patched_parts))  # unique
+
+            gh.issues.label_issue(api, settings.URN, pr_num, patched_parts)
 
             if is_approved:
                 __log.info("PR %d status: will be approved", pr_num)
--- a/settings.py
+++ b/settings.py
@@ -1,4 +1,5 @@
 from os.path import exists, abspath, dirname, join
+from collections import OrderedDict
 import misc
 import logging
 
@@ -90,6 +91,8 @@ REPO_LABELS = {
     "ci failed": "ff9800",
 
     "system": "8e44ad",
+    "issues": "4daf7c",
+    "pull_requests": "f2784b",
     "core": "c0392b",
     "web": "27ae60",
     "github api": "2c3e50",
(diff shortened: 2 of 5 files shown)


settings.py
```diff
@@ -126,3 +133,12 @@
 #     "password ": "chaos",
 #     "db ": "db"
 # }
+
+SOURCE_PARTS_NAME = {
+    "system": ["ansible/*", "etc/*", "memoize/*", "puppet/*", "startup.d/*", "chaos_wrapper.sh"],
+    "core": ["cron/*", "chaos.py"],
```
PlasmaPower: Maybe separate out files in `cron`, as they handle very different parts of chaos.

Swizz:
I love the way Chaos judge me with labels, and remove it after change. Lets add more labels !

settings.py
Swizz: Everything is possible. I use `fnmatch`
I worked on feature, that why this is in settings o/

Suggest me change you want :)

PlasmaPower:
```
"issues": ["cron/*issue*.py"],
"pull_requests": ["cron/poll_pull_requests.py"],
"core": ["cron/*", "chaos.py"],
```

settings.py
```diff
@@ -126,3 +133,12 @@
 #     "password ": "chaos",
 #     "db ": "db"
 # }
+
+SOURCE_PARTS_NAME = {
```
PlasmaPower: Should be an `OrderedDict`. Otherwise, some generalizations could end up before the specifics (e.g. what if `misc` ends up first?).

PlasmaPower (changes requested):

settings.py
Swizz: I am not a Python expert :/

OrderedDict ensure iteration to follow the initial given order ?
Dict iteration is unpredictable ?

PlasmaPower: Exactly

Swizz:
Here we go

PlasmaPower:
> Update settings **for meritocracy pleasure**

:disappointed:

PlasmaPower (approved):

cron/poll_pull_requests.py
```diff
@@ -95,6 +96,17 @@ def poll_pull_requests(api):
             # is our PR in voting window?
             in_window = gh.prs.is_pr_in_voting_window(api, pr, voting_window)
 
+            # Add the rights labels
+            __log.info("Adding PR %d labels regarding is patch", pr_num)
```
anythingbot, line 100: I think this should be changed to "Adding PR %d labels regarding its patch" unless you had something else in mind

cron/poll_pull_requests.py
```diff
@@ -95,6 +96,17 @@ def poll_pull_requests(api):
...
+
+            patched_files = gh.prs.get_patch(api, settings.URN, pr_num)
```
PlasmaPower, line 102: You might want to try-catch this (and the following labeling code) just in case.

o3LL:
Or extend meritocracy ?

andrewda:
@anythingbot There are pending reviews that still need to be addressed before approving the PR. @PlasmaPower suggested adding a try/catch block to prevent even more ChaosBot crashes, and there's a typo to fix that you reviewed.

PlasmaPower:
I'd prefer error handling before a positive review, but I guess I'll approve it anyway.

andrewda:
@PlasmaPower I think error handling is pretty important, and should probably be added first.

PlasmaPower:
Okay. I'll create an issue to get people's opinions on if we should approve in these cases.

andrewda:
@Swizz resubmit with the above fixes ❤️

Swizz:
It was around 3 AM in my timezone when you writed this.
Meritocrates are rude !

I will work on it.

hongaar:
Really nice feature!

Swizz:
Reopening this PR works or do I need to open a fresh new one ?

(Ow I cant, only Chaosbot can, so, lets go to a new one x) )

PR #515, opened 2017-06-04
Fully deprecate old web server and fix new one

mdcfe:
* Move API endpoints to `/api/`
* Reintroduce error page rendering
* Kill old web server, just in case it's still running
* Reload nginx when updating ChaosBot to ensure that configs are reloaded


andrewda (approved):
Awesome! 👏

PlasmaPower (approved):

andrewda:
Hmm... Doesn't seem to be working @md678685 

http://chaosthebot.com/api/voters

PR #542, opened 2017-06-09
Add a news link to term.js

dbpokorny:
Become a part of the Anythingbot Programming News Network today!

This PR adds a link to botwillacceptanything television---BWAA TV---which has just been remodeled.

If you want to change the video feed, or have any other suggestions, please make an issue or pull request for anythingbot/anythingbot.

### DIFF
--- a/server/static/js/term.js
+++ b/server/static/js/term.js
@@ -4,13 +4,16 @@ let terminal;
 
 $("#term").terminal({
     help: () => {
-        terminal.echo("github, gitter, help");
+        terminal.echo("github, gitter, news, help");
     },
     github: () => {
         terminal.echo("<a href='https://github.com/chaosbot/Chaos'>Click here</a> to visit the GitHub repo.", { raw: true });
     },
     gitter: () => {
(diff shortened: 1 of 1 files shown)


andrewda:
A little confused here —anythingbot is a different project, right...?

dbpokorny:
@andrewda yes, another democracy bot

Edit: well, actually it is technically a democracy of 1 (dictatorship) but it comes from botwillacceptanything, a democracy bot which is no longer operational

andrewda:
So why are we adding it to this project? We could add our own television system, I guess, but it seems a little weird to link to anythingbot without explanation.

dbpokorny:
@andrewda There are many reasons to add a link, for example the promotion of ties between the projects. I would be happy to explain further, for example by giving more reasons.

dbpokorny:
Would you change your vote if someone made a video of an introduction and comparison / contrast between the chaosbot and botwillacceptanything? This could start as an essay rendered as an audio or video project, perhaps with some NPR-inspired theme music.

PR #543, opened 2017-06-10
Reopen #443 according to #364

eamanu:
Fix some bugs
Add new twitter post


PlasmaPower:
You forgot to add `privkey` to the gitignore. You could also only use `privkey` when testing, or when `/etc/privkey` cannot be opened.

eamanu:
Thanks of GOD `Travis build passed`

encryption.py
```diff
@@ -45,4 +45,4 @@ def decrypt(ciphertext):
     return decrypt
 
 
-decrypt = create_decryptor("/etc/privkey", "server/pubkey.txt")
+decrypt = create_decryptor("etc/privkey", "server/pubkey.txt")
```
PlasmaPower, line 48: Still not `.gitignore`d

PlasmaPower (changes requested):

PR #545, opened 2017-06-10
Fix "ci failed" issue v2

PlasmaPower:
Previously, chaosbot would remove the tag when the build was still pending. That caused problems, as seen in #543. With this, it should be stable (it won't be removed then added again).


PlasmaPower:
WIP-ing as it's night time in a lot of places so it'll be practically impossible to get passed.

mark-i-m (approved):

PlasmaPower:
Repushing and un-WIP-ing to give it a normal voting window.

mark-i-m (approved):

PR #549, opened 2017-06-10
Relax ESLint rules

qgustavor:
Those are too restrict. It's not good for chaos. **We need chaos.**
Changed airbnb to eslint:recommended, which is not too strict.
As documented [here](http://eslint.org/docs/rules/) it "report common problems" instead of enforcing a specific code style.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ python:
 install:
   - "nvm install node"
   - "nvm use node"
-  - "npm install -g eslint@^3.19.0 eslint-plugin-jquery@latest eslint-plugin-jsx-a11y@^5.0.1 eslint-plugin-import@^2.2.0 eslint-plugin-react@^7.0.1 eslint-config-airbnb"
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -4,7 +4,7 @@ python:
 install:
   - "nvm install node"
   - "nvm use node"
-  - "npm install -g eslint@^3.19.0 eslint-plugin-jquery@latest eslint-plugin-jsx-a11y@^5.0.1 eslint-plugin-import@^2.2.0 eslint-plugin-react@^7.0.1 eslint-config-airbnb"
+  - "npm install -g eslint@^3.19.0 eslint-plugin-jquery@latest eslint-plugin-jsx-a11y@^5.0.1 eslint-plugin-import@^2.2.0 eslint-plugin-react@^7.0.1"
```
mdcfe: The JSX and React plugins are only installed because they are dependencies of the Airbnb config.

mdcfe (changes requested):

PR #550, opened 2017-06-10
Re Reopen #443 according to #364 - Twitter Api Working

eamanu:
Fix encryptation problems, regenerate twitter keys @PlasmaPower


twitter_api/Twitter.py
```diff
@@ -0,0 +1,7 @@
+def PostTwitter(message, api_twitter):
+    if len(message) > 140:
+        print('Post has more of 140 chars')
+        return 1
+    api = api_twitter
+    api.PostUpdate(message)
```
PlasmaPower:
```py
try:
  api.PostUpdate(message)
except:
  log.exception("Failed to post to Twitter")
```

Just in case an unexpected issue occurs, or there's an encoding problem.

eamanu: Which module I've to import to use `log.exception()` ?

kylerchin (approved):

kylerchin:
:100: Looking good! Any account name yet?

kylerchin:
I can give you "realChaosBot" on twitter if you would like. ;)

eamanu:
@kylerschin  https://twitter.com/ChaobotsC

twitter_api/Twitter.py
PlasmaPower: At the beginning of the module:

```py
import logging

log = logging.getLogger("twitter")
```

eamanu: ready

cron/poll_pull_requests.py
```diff
@@ -94,6 +97,12 @@ def poll_pull_requests(api):
             if is_approved:
                 __log.info("PR %d status: will be approved", pr_num)
 
+                # Make a tweet
+                message_twitter = datetime.datetime.ctime(
+                    datetime.datetime.now()) +\
+                    " - PR {pr_num} status: will be approved".format(pr_num=pr_num)
```
PlasmaPower: This will spam the twitter feed, as it will happen every 30 seconds for every PR on its way to being approved. Either only do it once for every PR, or eliminate it altogether.

eamanu: Ready

PlasmaPower (approved):
:tada:

dbpokorny:
@eamanu "encryptation" isn't a word, you mean "encryption".

@PlasmaPower are there plans to make the bot generate the encryption keys when it needs to by itself, without your supervision? I was under the impression that the encryption work was supposed to lead to that point eventually.

PlasmaPower:
@dbpokorny What do you mean? It already generates its own encryption keys.

dbpokorny:
@PlasmaPower oh dear, I thought @eamanu was asking you to regenerate the twitter keys and somehow send them to the bot. My misunderstanding.

PlasmaPower:
@dbpokorny No, it was the reverse - I asked @eamanu to regenerate the twitter keys. :smile:

eamanu:
@dbpokorny I write 'encryptation' repeatedly... My fouls. Sorry for my spanglish :laughing:

mdcfe:
This breaks dev environments almost completely, as ChaosBot tries to decrypt the Twitter token with the locally-generated private key, throws an error because it's the wrong key and crashes.

